PR #2, opened 2023-10-18
docs: Manually create Equinix cluster

rossf7:
Adds docs for manually creating a cluster with CAPI / CAPEM to try things out while we work on the pipeline design.

I've added comments for decisions I'm not sure on or that are worth highlighting.

One thing to note is the Kubernetes API will be public. As we've discussed as a later step we could make it private and use a GitOps tool like Flux to manage the components running in the cluster.

### DIFF
--- a/docs/create_equinix_cluster.md
+++ b/docs/create_equinix_cluster.md
@@ -1,0 +1,109 @@
+# Create Equinix Metal cluster using CAPI / CAPEM
+
+## Overview
+
+- We are using machines contributed to the CNCF by Equinix Metal.
+- These steps manually create a cluster using CAPI / CAPEM using a local
+Kind cluster for the management cluster.
+- These docs are based on the Equinix CAPI [guide](https://deploy.equinix.com/developers/guides/kubernetes-cluster-api/).
+- Using [Podman](https://github.com/kubernetes-sigs/cluster-api-provider-packet/issues/496)
+is recommended by Equinix for MacOS.
+- We plan to automate these steps later using GitHub Actions and an IaC tool.
+
+## Create Cluster
+
+- Log in to the Equinix Metal [console](https://console.equinix.com/) and get
+the project API key from the project settings. If it doesn't exist create it.
+
+```sh
+export PACKET_API_KEY="<PROJECT_API_KEY>"
+```
+
+- Install the CAPI controllers using [clusterctl](https://cluster-api.sigs.k8s.io/user/quick-start.html#install-clusterctl).
+
+```sh
+clusterctl init --infrastructure packet
+```
+
+- Set env vars with cluster config.
+
+```sh
+# Get the project ID from the project settings in the console
+export PROJECT_ID="<PROJECT_ID>"
+
+# Use Paris metro (Equinix region)
+export METRO="pa"
+
+# Use Ubuntu 22.04 with cGroup v2 for Kepler
+export NODE_OS="ubuntu_22_04"
+
+# The pod and service CIDRs for the new cluster
+export POD_CIDR="192.168.0.0/16"
+export SERVICE_CIDR="172.26.0.0/16"
+
+# Use node type with Intel CPUs for RAPL support
+export CONTROLPLANE_NODE_TYPE="m3.small.x86"
+export WORKER_NODE_TYPE="m3.small.x86"
+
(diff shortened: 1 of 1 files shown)


docs/create_equinix_cluster.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+- Set env vars with cluster config.
+
+```sh
+# Get the project ID from the project settings in the console
+export PROJECT_ID="<PROJECT_ID>"
+
+# Use Paris metro (Equinix region)
+export METRO="pa"
```
rossf7, line 35: Do we want to consider carbon intensity when deciding which metro to use?

I chose Paris since France has a low average carbon intensity and this metro has a good selection of node types.

Full list is here and we can easily change if I'm over thinking this :)
https://deploy.equinix.com/developers/docs/metal/locations/metros/

docs/create_equinix_cluster.md
```diff
@@ -0,0 +1,109 @@
...
+# Use Ubuntu 22.04 with cGroup v2 for Kepler
+export NODE_OS="ubuntu_22_04"
+
+# The pod and service CIDRs for the new cluster
+export POD_CIDR="192.168.0.0/16"
+export SERVICE_CIDR="172.26.0.0/16"
+
+# Use node type with Intel CPUs for RAPL support
+export CONTROLPLANE_NODE_TYPE="m3.small.x86"
+export WORKER_NODE_TYPE="m3.small.x86"
```
rossf7, line 46: I think its safest to use a node type with an Intel CPU. I tested with an AMD CPU and Kepler couldn't access RAPL.

Screenshot 2023-10-18 at 11 15 34

I think this node type have good enough resources for now. WDYT?

See https://deploy.equinix.com/product/bare-metal/servers/

docs/create_equinix_cluster.md
```diff
@@ -0,0 +1,109 @@
...
+
+- Generate cluster manifests.
+
+```sh
+clusterctl generate cluster wg-green-reviews \
+  --kubernetes-version v1.28.2 \
+  --control-plane-machine-count=1 \
+  --worker-machine-count=1 \
+  > wg-green-reviews.yaml
+```
```
rossf7, line 60: Cluster will be named `wg-green-reviews` and I went with the latest release of Kubernetes.

docs/create_equinix_cluster.md
```diff
@@ -0,0 +1,109 @@
...
+
+- SSH to each cluster node and ensure Kepler dependencies are installed
+(user is named `root`).
+
+```sh
+apt update
+apt install linux-headers-$(uname -r)
+apt install linux-modules-extra-$(uname -r) 
+modprobe intel_rapl_common
+```
```
rossf7, line 95: I didn't actually need to run these steps but I think its safest to include them since they are requirements for Kepler. WDYT?

docs/create_equinix_cluster.md
nikimanoledaki: This is a very important callout! We can hardcode Paris for now and make the testing carbon-aware in the future. Would you like to create a new issue for this to add it to our backlog? 🌟

docs/create_equinix_cluster.md
nikimanoledaki: +1 for using an Intel CPU and starting with `small`! 👍

docs/create_equinix_cluster.md
nikimanoledaki: WDYT about also exporting the kubernetes version as an env var to surface it and make it easier to update in the future?

docs/create_equinix_cluster.md
nikimanoledaki: Good to validate that the kernel headers, I always double-check this as well 👍

nikimanoledaki (approved):
Wonderful, thank you @rossf7! 🌟 🎉

docs/create_equinix_cluster.md
rossf7: Yes sure I'll create an issue for it. 👍

docs/create_equinix_cluster.md
rossf7: Yes good idea! I've added that.

Issue #1, opened 2023-10-04
[Action] Bootstrap Kubernetes cluster with IaC tooling


nikimanoledaki:
## Cluster API
We may want to use the [Equinix Metal Cluster API Provider (CAPEM)](https://deploy.equinix.com/developers/guides/kubernetes-cluster-api/) for our cluster bootstrapping on the community cluster. Alternatives such as Ansible or Firecracker microVMs are being considered, to work with Falco's setup: https://github.com/cncf/tag-env-sustainability/issues/182

## Requirements
The cluster requirements are listed in the [design doc](https://docs.google.com/document/d/19fzZW-IMv2kDNatKFHeHh7wqcEN0e2N60wzxvCGZd48/edit#heading=h.o8b3llcvsfjw).

## Equinix infrastructure access
This issue will help us know more about the kind of access that will be needed for individual contributors to the infra. Please see [this](https://github.com/cncf/tag-env-sustainability/issues/220#issuecomment-1739278591) for some of the available options, and follow up in that thread with any questions/issues.

## Documentation

We should [document](https://docs.google.com/document/d/19fzZW-IMv2kDNatKFHeHh7wqcEN0e2N60wzxvCGZd48/edit#heading=h.o8b3llcvsfjw) this process as we go. 

## Development environment

Dev environment setup tracked in this issue: https://github.com/cncf-tags/green-reviews-tooling/issues/3

rossf7:
@nikimanoledaki I've been investigating using CAPI  / CAPEM and I think us not having access to a permanent management cluster is a problem.

In the Equinix docs they have an alternative approach intended for management clusters using K3s managed by Pulumi.
https://deploy.equinix.com/developers/guides/k3s-management-plane/

I've added more detail in the design [doc](https://docs.google.com/document/d/19fzZW-IMv2kDNatKFHeHh7wqcEN0e2N60wzxvCGZd48/edit). PTAL

Do you think this is a good direction?

nikimanoledaki:
@rossf7 and I discussed this and wrote a summary [here](https://github.com/cncf/tag-env-sustainability/issues/182#issuecomment-1752902159). We will have more information after today's WG Green Reviews meeting with the Falco maintainers.

rossf7:
Update following the WG meeting and discussions I had since with @nikimanoledaki 

We think an important factor is how we isolate between test runs to have accurate results.

- Least isolation needed would be to run each test on a separate node
[...]

PR #6, opened 2023-11-07
Provision cluster and bootstrap flux

rossf7:
This is provisions the Equinix infra using OpenTofu and K3s and bootstraps Flux.

These secrets need to be added to the repo.

- `AWS_ACCESS_KEY_ID` for S3 bucket to store state
- `AWS_SECRET_ACCESS_KEY`
- `EQUINIX_AUTH_TOKEN` the Equinix project API key
- `EQUINIX_PROJECT_ID` the Equinix project ID
- `FLUX_GITHUB_TOKEN` the GitHub token used to bootstrap Flux
- `K3S_TOKEN` the K3s token for joining nodes 
- `SSH_PUBLIC_KEY` added as an Equinix project SSH key
- `SSH_PRIVATE_KEY` needed so tofu can access nodes

## Issue

- Closes https://github.com/cncf-tags/green-reviews-tooling/issues/1


infrastructure/equinix-metal/ansible-playbooks/main-playbook.yml
```diff
@@ -0,0 +1,6 @@
+---
+# Main playbook run to provision cluster.
+# TODO Implement playbook
+- name: Empty Play
+  hosts: all
+  tasks: []
```
rossf7, line 6: Just an empty playbook to confirm Ansible can access the server.

infrastructure/equinix-metal/main.tf
```diff
@@ -0,0 +1,61 @@
...
+  EOF
+
+  depends_on = [
+    equinix_metal_device.control_plane
+  ]
+}
+
+resource "null_resource" "ansible_playbook" {
+  triggers = {
+    always_run = "${timestamp()}"
```
rossf7, line 53: Right now the Ansible playbook is run each time. Do you think this is OK? Or can we detect if the playbook has changed?

dipankardas011: - If then infrastructure changes then configuration needs to be applied
- If no change in infrastructure and we are applying will it affect existing cluster?

May be this conditional statement can help
```
If (infrastructure changes)
  Apply scripts
Else 
  If (scripts changed)
    Apply scripts
```

dipankardas011: Not really sure, never used null_resources
but from few projects they use `null_resources` for executing some commands
and for user_data its sepecific to VM specific first time boot scripts

dipankardas011: need to check it out

dipankardas011: refering: https://spacelift.io/blog/terraform-provisioners#how-to-use-terraform-provisioners

we can use the remote exec to get the k3s token
and then we can use the out put from this to insert in another resource for (VM for workernodes) which have to wait for both the controlplane node with k3s installed and the k3s token

dipankardas011: few projects they use null_resources for executing some commands
and for user_data its sepecific to VM specific first time boot scripts

dipankardas011: Also we can refer to this https://docs.k3s.io/installation/airgap#installing-k3s-in-an-air-gapped-environment

we can do is create a random string using terraform resource and use that as the k3s token

by this we eliminate the use of custom commands to be executed and we can use that token to join other workernodes

dipankardas011: and for installing cilium we can just pass the flag for nocni and then we can install cilium using helm install

rossf7: > few projects they use null_resources for executing some commands
> and for user_data its specific to VM specific first time boot scripts

Yes, I used it for triggering Ansible but I wasn't sure how widely used it is and better to use user_data. I'll remove.

> we can do is create a random string using terraform resource and use that as the k3s token

Nice! I like the idea of generating the k3s token.

dipankardas011: Can I open a PR for `rossf7:wip-opentofu-spike` add some scripts there for k3s?
or we should wait for everyone opnions?

rossf7: @dipankardas011 Thanks if you can open a PR with the scripts I think that would be helpful. So we can see what it would look like using k3s.

We do need to get more opinions and discuss the tools at the next WG meeting. I've added a note to the PR description that the tools used are provisional and still to be agreed.

dipankardas011: okay np I will wait for more opnions then I will create PR for the scripts
to avoid dual work 😄

rossf7: Good call 👍 better to wait till we get more feedback

dipankardas011: till now Here is what it looks like
```tf
resource "random_string" {
   // some value
}

resource "controlplanevm" {

  user_data = << EOF
  #!/bin/bash
  curl -sfL https://get.k3s.io | INSTALL_K3S_CHANNEL="1.28-..." K3S_TOKEN=${random_string.output} sh -s - server \
	--node-taint CriticalAddonsOnly=true:NoExecute \
	--flannel-backend=none \
	--disable-network-policy \
	--tls-san %s
  EOF
}

resource "WorkerNodesvm" {
  for loop for 1 .. N

  depends_on = [ controlplanevm ]

  user_data = << EOF
  #!/bin/bash
  curl -sfL https://get.k3s.io | INSTALL_K3S_CHANNEL="1.28-..." sh -s - agent --token ${random_string.output} --server https://${controlplanevm.networks.ip}:6443
  EOF
}
```

dipankardas011: one issue is how to know if the controlplane is configured before we execute the worknodes creation step
one solution which come to my mind is use another resource which checks if k3s systemd service is op and running

rossf7: @dipankardas011 that looks good and a null resource to check k3s is up on the control plane node makes sense to me.

I just pushed some small changes to remove Ansible and create worker nodes based on a list of node names.

leonardpahlke:
@nikimanoledaki has access to add repo secrets

infrastructure/equinix-metal/main.tf
dipankardas011: nice

nikimanoledaki:
I'm going to pre-empt a review comment already - could you [enable GPG so that signed commits are verified](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification), please? Thank you!

dipankardas011:
I am using gpg keys but not sure why its not giving a verified tag in every commit?

nikimanoledaki:
Yes, nearly there! 🤞 Ross, cp sounds good 👍 
Could it be a mismatch with the email address used to sign the commit vs the one used for the gpg key? 🤔

infrastructure/equinix-metal/.terraform.lock.hcl
dipankardas011, line 1: This file should be present ?

dipankardas011 (approved):
Nice ready for merge 👍🏼

infrastructure/equinix-metal/.terraform.lock.hcl
rossf7: > Terraform automatically creates or updates the dependency lock file each time you run [the terraform init command](https://developer.hashicorp.com/terraform/cli/commands/init). You should include this file in your version control repository so that you can discuss potential changes to your external dependencies via code review, just as you would discuss potential changes to your configuration itself.

https://developer.hashicorp.com/terraform/language/files/dependency-lock

Yes, Terraform docs say it should be committed to ensure providers are in sync. I couldn't find the equivalent docs for OpenTofu but it should work the same.

rossf7: 🤔  Although we use `tofu init -upgrade` which will install the latest version of plugins.

> -upgrade Upgrade all previously-selected plugins to the newest version that complies with the configuration's version constraints. This will cause OpenTofu to ignore any selections recorded in the dependency lock file, and to take the newest available version matching the configured version constraints.

https://opentofu.org/docs/cli/commands/init#plugin-installation

When I get a chance I'll try removing it and see what the effects are.

rossf7: We don't need the lock file as we specify the provider versions we need. 

So it's gone 🎉

rossf7:
I made some "final final" changes and renamed some of the vars. The PR description is updated with the new secret names.

Just needs the S3 bucket and then the secrets can be added and this can be merged.

FYI @nikimanoledaki @dipankardas011

rossf7:
Going to create a fresh branch and cp to clean the history a bit. Please ignore for now. 🙏

rossf7:
Closing in favor of #28

PR #24, opened 2024-01-12
Delete additional flux initialisation

nikimanoledaki:
Just a small PR to delete the additional flux directory from a second initialisation! 

cc @antoniodituri


rossf7 (approved):
lgtm

PR #26, opened 2024-01-16
Use kube-prometheus-stack chart

rossf7:


clusters/kepler.yaml
```diff
@@ -32,3 +32,5 @@ spec:
       sourceRef:
         kind: HelmRepository
         name: kepler
+  dependsOn:
+    - name: kube-prometheus-stack
```
rossf7, line 36: Adds the dependency on the service monitor CRD

clusters/flux-system/gotk-sync.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 spec:
   interval: 1m0s
   ref:
-    branch: main
+    branch: prometheus-stack
```
rossf7, line 11: I couldn't find a way to keep prometheus deployed without switching the branch 😅 

```
k get helmrelease -A
NAMESPACE     NAME                    AGE   READY   STATUS
flux-system   kepler                  4d    True    Release reconciliation succeeded
flux-system   kube-prometheus-stack   90s   True    Release reconciliation succeeded
```

Could you revert this to main before merging?

nikimanoledaki: Sure! Will remember to revert the commit/change to point at `main` 👍

nikimanoledaki (approved):
Flux changes LGTM 👍

clusters/kube-prometheus-stack.yaml
```diff
@@ -11,17 +11,19 @@ spec:
...
 kind: HelmRelease
 metadata:
-  name: prometheus
-  namespace: monitoring
+  name: kube-prometheus-stack
+  namespace: flux-system
 spec:
   interval: 24h
+  install:
+    createNamespace: true
```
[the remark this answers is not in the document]
nikimanoledaki: 💯

clusters/flux-system/gotk-sync.yaml
nikimanoledaki: Switching branches gave me an idea: we could use a branch strategy to test the different CNCF projects when we need to deploy them with Flux. 💡 

For example, we could move all the Falco-specific config (mainly the flux config for the synthetic workload and Falco itself) to its own `falco` branch. So, `main` contains all the basic components, and the branches contain specific components. 

Not something for now though, we can revisit this later, but wanted to share the idea 😄

rossf7: I like this idea and the PR based workflow for the projects is nice.

Eventually with a lot of projects branch management might get hard. Maybe at that point we could suspend the resources and then un-suspend when we want to deploy them? 🤔

PR #28, opened 2024-01-19
Provision cluster and bootstrap flux

rossf7:
This is provisions the Equinix infra using OpenTofu and K3s. It then installs Cilium for CNI and bootstraps Flux which is used to manage the rest of the components.

## Issue

- Closes https://github.com/cncf-tags/green-reviews-tooling/issues/1


infrastructure/equinix-metal/main.tf
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  provisioner "file" {
+    content = <<EOF
+GITHUB_TOKEN=${var.flux_github_token}
+KUBECONFIG=/etc/rancher/k3s/k3s.yaml
+EOF
+    destination = "/tmp/flux_env_vars"
+  }
+}
```
dipankardas011, line 136: @rossf7 if we are copying the kubeconfig from the controlplane then we can rename resource block to `get-kubeconfig`?

also should we use this kubeconfig to install kepler and prometheus-stack with grafana dashboard or will plan it in some other workflow?

infrastructure/equinix-metal/main.tf
```diff
@@ -0,0 +1,157 @@
...
+
+  provisioner "remote-exec" {
+    inline = [
+      "curl -sfL https://get.k3s.io | INSTALL_K3S_CHANNEL=${var.k3s_version} K3S_TOKEN=${var.k3s_token} sh -s - server --node-taint node-role.kubernetes.io/control-plane=:NoSchedule --flannel-backend=none --disable-network-policy --disable=traefik",
+      "systemctl is-active --quiet k3s.service",
+    ]
+  }
+}
+
+// NOTE: to extract KUBECONFIG we need to copy kubeconfig in controlplane:/etc/rancher/k3s/k3s.yaml
```
dipankardas011, line 62:
```suggestion

```

if we have implemented it we should remove it

dipankardas011 (commented):
Had some doubts

infrastructure/equinix-metal/main.tf
rossf7: @dipankardas011 The separate block is mainly for the github token. This allows us to view logs from the `flux bootstrap` command. Otherwise tofu hides them to avoid outputting the token.

For installing kepler and prometheus as far as possible we should apply all changes via flux. Flux is granted `cluster-admin` privileges as part of the bootstrap process. https://github.com/cncf-tags/green-reviews-tooling/blob/86f136fb88e77a6743dcec6acc4ca83e1d484f7c/clusters/flux-system/gotk-components.yaml#L247

We still need the kubeconfig for manual trouble shooting but anyone with ssh access can get it and generate a readonly kubeconfig using the steps you're working on.

infrastructure/equinix-metal/main.tf
rossf7: Good catch. I've removed it.

infrastructure/equinix-metal/main.tf
dipankardas011: got it thanks @rossf7  for clarifying :)

dipankardas011 (approved):
LGTM 🚀

nikimanoledaki (approved):

PR #29, opened 2024-01-23
Create monitoring namespace

rossf7:
Fix for error we hit in https://github.com/cncf-tags/green-reviews-tooling/actions/runs/7628160767

The kepler dashboard configmap is created by Flux so we need to create the namespace.

### DIFF
--- a/clusters/base/kube-prometheus-stack.yaml
+++ b/clusters/base/kube-prometheus-stack.yaml
@@ -15,8 +15,6 @@ metadata:
   namespace: flux-system
 spec:
   interval: 24h
-  install:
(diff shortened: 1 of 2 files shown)


clusters/base/kube-prometheus-stack.yaml
```diff
@@ -15,8 +15,6 @@ metadata:
   namespace: flux-system
 spec:
   interval: 24h
-  install:
-    createNamespace: true
```
rossf7, line 19: Removing this since the configmap is created before the helmrelease CR

nikimanoledaki (approved):

PR #21, opened 2024-01-11
new(falco): add flux deployment config

incertum:
Add Falco flux deployment config.

@nikimanoledaki first time using Flux. Looking forward to hearing more tips from you and happy to make any necessary adjustments.

Our new dedicated repo contains relevant docs https://github.com/falcosecurity/cncf-green-review-testing

Action Items:
- [ ] Clarify cluster requirements and setup: https://github.com/falcosecurity/cncf-green-review-testing/issues/2
- [ ] Synthetic Workloads for Falco https://github.com/falcosecurity/cncf-green-review-testing/issues/3, also checkout our current setup in https://github.com/falcosecurity/cncf-green-review-testing. This is just to get us going with something. Open to make any adjustments or change the approach altogether.

Tracking:
- https://github.com/falcosecurity/cncf-green-review-testing/issues/4

CC Falco maintainers @leogr @LucaGuerra @maxgio92

PR #32, opened 2024-01-24
Move falco gitrepo CR to falco namespace

rossf7:
Follow up to #21 

Reconciling the kustomization failed because the gitrepo was in the flux-system namespace.

@nikimanoledaki Should we keep both resources in the falco ns? I'm not sure what the best practice is here.

### DIFF
--- a/clusters/projects/falco/falco.yaml
+++ b/clusters/projects/falco/falco.yaml
@@ -6,7 +6,7 @@ apiVersion: source.toolkit.fluxcd.io/v1
 kind: GitRepository
 metadata:
   name: falco-cncf-green-review-testing
-  namespace: flux-system
+  namespace: falco
 spec:
   interval: 6h
   ref:


nikimanoledaki:
Looking at the Flux docs for [GitRepository](https://fluxcd.io/flux/components/source/gitrepositories/#spec-examples), it mentions that Flux can support multiple so that shouldn't be the issue. `flux-system` should be the right namespace. 

I'm wondering if Flux needs to be bootstrapped a second time, this time to be able to watch the `falcosecurity/cncf-green-review-testing` repository:
```
flux bootstrap github --owner=$GITHUB_USER --repository=https://github.com/falcosecurity/cncf-green-review-testing --path=./kustomize
```

rossf7:
> Did changing the GitRepository namespace work for you

Yes when I created the GitRepository manually the changes were applied.

> I'm wondering if Flux needs to be bootstrapped a second time, this time to be able to watch the falcosecurity/cncf-green-review-testing repository:

That's a good shout. I'll do some more investigating on this.

nikimanoledaki (approved):
LGTM since this worked for you! 👍

PR #44, opened 2024-02-05
feat: added nodeSelector based on k8s node labels for `kube-prometheus`, `kepler`

dipankardas011:
## why

it bring about the

Issue #30, opened 2024-01-23
[Action] Separate nodes for Falco and our internal stack


rossf7:
This issue is to create separate kubernetes nodes for our internal stack (Flux / Prometheus) and Falco (first project we're measuring).

Node isolation is important to ensure we can measure the footprint of projects accurately.

See https://github.com/falcosecurity/cncf-green-review-testing/issues/2 for Falco node requirements

## Node requirements

The nodes will be managed by `opentofu` and have these names and node labels.

- green-reviews-worker-internal1

```
cncf-project: wg-green-reviews
cncf-project-sub: internal
```
- green-reviews-worker-falco1

```
cncf-project: falco
cncf-project-sub: falco-driver-modern-ebpf
```

We will start with using node labels and selectors for placing pods on nodes ~~but we may also need to introduce node taints and tolerations~~.

dipankardas011:
I think I can help

rossf7:
@dipankardas011 Thank you! My initial thought on this was to replace the list var of worker nodes [here](https://github.com/cncf-tags/green-reviews-tooling/blob/5f6a2ea6405ac0ce16a257b6432fceabf4e419e6/infrastructure/equinix-metal/variables.tf#L106) with a map that includes the labels. WDYT?

We would need to use node selectors for Prometheus, Flux and any other components so they run on the internal node. 

Another approach is to add a taint to the falco node and ask the falco team to add a toleration in https://github.com/falcosecurity/cncf-green-review-testing

cc @nikimanoledaki @AntonioDiTuri

rossf7:
@dipankardas011 That's a nice approach and in future we may want to run some workloads on our control plane node if we start to max out the "system" node where we run Prometheus, Flux etc.

A downside I see with a taint per project is we need to run Kepler on all nodes. So we'd need to add tolerations for all the taints.

How about we start with adding node selectors to the kube-prometheus-stack helm release and the flux bootstrap?

https://fluxcd.io/flux/installation/configuration/boostrap-customization/

If adding a node selector for each component becomes too hard to manage we can look at alternatives later.

dipankardas011:
## Query
1. I need to modify the fluxcd manifest for installing prometheus, kepler, ...
2. even if we do that prometheus exporter will be a daemonset thus present in every node, not sure with kepler!

rossf7:
> I need to modify the fluxcd manifest for installing prometheus, kepler, ...
[...]

PR #56, opened 2024-02-13
feat: Make node labels and plan configurable

rossf7:
#### What type of PR is this?

kind/feature

- Makes node labels configurable so we can run Falco and monitoring stack on different nodes.
- Makes Equinix metal plan (their name for instance type) configurable so we can run larger nodes when needed

https://deploy.equinix.com/developers/docs/metal/hardware/standard-servers/

Towards https://github.com/cncf-tags/green-reviews-tooling/issues/30 and unblocks https://github.com/cncf-tags/green-reviews-tooling/pull/44


nikimanoledaki (approved):
Great work @rossf7, thank you! 🙇

PR #59, opened 2024-02-16
Temporarily remove microservice demo from Falco namespace

nikimanoledaki:


raymundovr:
Just noticed this, as the Falco team would like to keep it,  I've asked about it [in here](https://github.com/falcosecurity/cncf-green-review-testing/issues/16#issuecomment-1950000853).

Issue #58, opened 2024-02-16
[Tracking] Bootstrap self-hosted GitHub Action runners


nikimanoledaki:
This issue tracks the configuration of GitHub Actions self-hosted runners with the Action Runners Controller (ARC) (see [step-by-step instructions](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/quickstart-for-actions-runner-controller)).

#### WG Green Reviews
_The following must be completed to create an idle test._
- [ ] Deploy ARC by adding Helm Chart via Flux
- [ ] [Authenticate to the GitHub API via ARC](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/authenticating-to-the-github-api)
- [ ] Configure a [runner scale set](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/quickstart-for-actions-runner-controller#configuring-a-runner-scale-set)

#### Falco Maintainers
_The following must be completed to create the first active test._
- [ ] Generate repository-level GitHub PAT https://github.com/falcosecurity/cncf-green-review-testing/issues/16
- [ ] Migrate any [load test](https://github.com/falcosecurity/cncf-green-review-testing/blob/main/kustomize/falco-driver/kmod/stress-ng.yaml) to a GitHub Actions workflow
- [ ] Store upstream GitHub PATs as Secrets

PR #63, opened 2024-02-19
feat: Add GitHub Actions self hosted runner

rossf7:
#### What type of PR is this?

kind/feature

Installs GitHub Actions Runner Controller https://github.com/actions/actions-runner-controller with a self hosted runner called `green-reviews-runner` for us to test with. 

Also configures SOPS to encrypt the k8s secret `green-reviews-pat` that stores the GitHub PAT needed by ARC to access the GitHub API.

For SOPS the GPG key used is stored in a secret in the flux-system namespace. For now I created this manually but later we should add this step to the tofu automation.

Towards https://github.com/cncf-tags/green-reviews-tooling/issues/58

This adds a new namespace called `greenreviews` for the runner scale set and the pods it creates.

My thinking is we will configure another runner scale set in the `falco` namespace for the Falco team but with the pods running on our internal node.

This is totally up for discussion! So if you think there is a better design please suggest it.


.github/workflows/arc_test.yaml
```diff
@@ -0,0 +1,11 @@
...
+on:
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: green-reviews-runner
+    steps:
+    - run: echo "🎉 This job uses runner scale set runners!"
+    - run: sleep 30
+    - run: echo "DONE"
```
leonardpahlke, line 11: is this job a placeholder?

clusters/.sops.pub.asc
leonardpahlke, line 1: perhaps we want to place this as a secret into the repo.

.github/workflows/arc_test.yaml
leonardpahlke: how can we test that the changes in this PR work?

clusters/.sops.yaml
```diff
@@ -0,0 +1,4 @@
+creation_rules:
+  - path_regex: .*.yaml
+    encrypted_regex: ^(data|stringData)$
+    pgp: BFAD1CA2762581D260113596323CD3AC9A04F637
```
leonardpahlke, line 4: where is this coming from? is there a CLI to generate this rnd str? do we want to store this as an ENV?

leonardpahlke: is it the GPG key?

leonardpahlke (commented):
👋 took a quick look. Have some questions about the PR. Also asked in Slack why we need a self hosted GitHub actions runner (and if there is not a simpler way).

rossf7:
> took a quick look. Have some questions about the PR. Also asked in Slack why we need a self hosted GitHub actions runner (and if there is not a simpler way).

@leonardpahlke I've set this back to draft while we discuss self hosted runners. I think there are some benefits over regular runners for our use case. 

Like we discussed we'll use the docs PR @nikimanoledaki is working on to do that.

rossf7:
We're going to start with GitHub public runners since the cluster k8s API is public and to have less to maintain.

We can come back to this later if we need self hosted runners or encrypting secrets for Flux with SOPS

PR #76, opened 2024-03-13
Add PUE to SCI dashboard

raymundovr:
#### What type of PR is this?
kind/documentation
kind/enhancement

GSF SCI specification states that cloud computing operations should include the efficiency of the data center where the software runs in the calculation, [see here](https://sci.greensoftware.foundation/#software-boundary). This efficiency is measured by the Power Usage Effectiveness (PUE).

This PR adds the PUE to the calculation shown in the dashboard and makes some additions to the documentation to include information about it.

None


raymundovr:
Hi @nikimanoledaki @rossf7 this is ready for review :)

rossf7 (commented):
@raymundovr Awesome, thanks for adding

Left some minor comments but then looks good. We'll most likely merge this from the 🚆 on Monday :D

raymundovr:
@rossf7 Thank you for checking this, all resolved. Sorry for the multiple commits, was trying to resolve using the Github tools but forgot about the sign-off.

clusters/base/falco-sci.yaml
```diff
@@ -167,7 +166,7 @@ data:
               "showLineNumbers": false,
               "showMiniMap": false
             },
-            "content": "\n\n\n# [SCI Equation = ](https://sci-guide.greensoftware.foundation/)\n# [(E * I) + M per R](https://sci-guide.greensoftware.foundation/)",
+            "content": "\n\n\n# [SCI Equation = ](https://sci-guide.greensoftware.foundation/)\n# [(E * I * PUE) + M per R](https://sci-guide.greensoftware.foundation/)",
```
nikimanoledaki, line 169: I would revert this to the top-level SCI formula `(E * I)`

docs/measurement/sci.md
```diff
@@ -10,6 +10,16 @@ TBC
 
 TBC
 
+## Power Usage Effectiveness (PUE)
```
nikimanoledaki, line 13:
```suggestion
### Power Usage Effectiveness (PUE)
```

The PUE then becomes a subsection of `E`

docs/measurement/sci.md
```diff
@@ -1,6 +1,6 @@
 # Software Carbon Intensity
 
-SCI = ([E](https://sci-guide.greensoftware.foundation/E) * [I](https://sci-guide.greensoftware.foundation/I)) + [M](https://sci-guide.greensoftware.foundation/M) per [R](https://sci-guide.greensoftware.foundation/R)
+SCI = ([E](https://sci-guide.greensoftware.foundation/E) * [I](https://sci-guide.greensoftware.foundation/I) * [PUE](https://learn.greensoftware.foundation/energy-efficiency/#power-usage-effectiveness)) + [M](https://sci-guide.greensoftware.foundation/M) per [R](https://sci-guide.greensoftware.foundation/R)
```
nikimanoledaki, line 3: I would revert this as well to keep the top-level general SCI formula here. Then we can create a separate formula that showcases how we adapted the formula to the cloud-native scenario in our architecture. Something like this:

> Cloud-native SCI = ((CPU energy + memory energy) * PUE * static carbon coefficient) + embodied carbon of Equinix infrastructure 

Something like this 🤔 What do you think?

nikimanoledaki (changes requested):
This is great, thank you for adding the PUE, @raymundovr! Really good call 😄 

My only feedback is that I would make PUE part of `E` as this is also the guidance from the SCI documentation which you linked to:

> If the boundary includes on-premise and/or cloud data center operations, E should take into account the efficiency of the data center, including cooling and other energy consumption necessary to operate a data center. The data center's energy efficiency is usually available as a PUE (Power Usage Effectiveness) value.

Left a couple of comments to address this. I'm open to discussing this further, let me know what you think!

docs/measurement/sci.md
raymundovr: To be clear, here we'll replace this into `SCI = (E + M) per R`? and then, in a separate section, develop the `E` to showcase the `Cloud-native SCI`?
Would we then set this "whole" formula in the dashboard as well or only in the docs?
As we're using the `container_total` from Kepler, which includes more than CPU and memory, I'd label it as "container energy".

clusters/base/falco-sci.yaml
raymundovr: I was also hesitating on this, and at the end I went for it to avoid having multiple formulas on different places, also because people become aware of the components that contribute to the emissions calculation. I think that this also relates to [this comment](https://github.com/cncf-tags/green-reviews-tooling/pull/76/files/3a6266ceae99f40aaa367174ffb899385caf1d50#r1529137645)

raymundovr:
Hi @nikimanoledaki 
Noticed that in the docs the `PUE` was already removed from the equation. To align it, I went with the proposed change for the `falco-sci.yaml` as well.
This resolves with your suggested changes.

nikimanoledaki (approved):
@raymundovr thank you for your patience and apologies for the delay! LGTM 🎉 💚

PR #88, opened 2024-04-16
Proposal 1: Trigger and Deploy

### LINKED ISSUE #84: [ACTION] Proposal 1: Trigger and Deploy
## Task Description

- Parent Issue #83

### Goals to achieve
- [x] We first need to draft the proposal #88
- [ ] Find a way to trigger the execution of the GitHub Action on the Falco side needs to be defined (e.g. webhook) [Resource](https://docs.github.com/en/webhooks/webhook-events-and-payloads#release)
- [ ] Ask Falco to implement the trigger on their hand
- [ ] Define GH Action steps: Falco needs to be deployed via Flux
- [ ] Test the deployment via Falco trigger

dipankardas011:
#### What type of PR is this?
kind/documentation

Fixes #84

cc @rossf7


rossf7 (commented):
Good start 👍 first round of feedback from me

docs/proposals/proposal-trigger-and-deploy.md
```diff
@@ -73,14 +73,58 @@ we will then deploy the benchmarking workfload for the project, in this case its
...
+        echo "version ${{ github.event.inputs.version }}"
+```
+
+for invoking this
+
+```bash
+curl -X POST \
+     -H "Accept: application/vnd.github.v3+json" \
+     -H "Authorization: token $GITHUB_PAT" \
+     https://api.github.com/repos/rossf7/green-reviews-tooling/actions/workflows/trigger_test.yaml/dispatches \
```
rossf7, line 115:
```suggestion
     https://api.github.com/repos/cncf-tags/green-reviews-tooling/actions/workflows/pipeline.yml/dispatches \
```

 I'm proposing pipeline.yml as the name for the top level action. WDYT? 

Actions are composable so we could call sub workflows from there. Using .yml to follow github naming even though I prefer .yaml 😆

docs/proposals/proposal-trigger-and-deploy.md
```diff
@@ -73,14 +73,58 @@ we will then deploy the benchmarking workfload for the project, in this case its
...
+     -d '{"ref":"main", "inputs": {"cncf_project": "falco", "cncf_project_sub": "modern-ebpf","version":"0.37.0"}}'
```
rossf7, line 116: For the ref maybe we should use a tag here?

The docs on release management are pretty good

https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-release-management-for-actions

Although I think some recommendations like having the action in a separate repo are too heavy for our needs

dipankardas011: > Although I think some recommendations like having the action in a separate repo are too heavy for our needs

Yes 

> For the ref maybe we should use a tag here?

yes

docs/proposals/proposal-trigger-and-deploy.md
```diff
@@ -0,0 +1,247 @@
+# GRP-001 - Trigger and Deploy GitHub Action workflow from an upstream CNCF project
```
rossf7, line 1: @dipankardas011 Little nitpick but please rename the file to start `proposal-001-`

docs/proposals/proposal-trigger-and-deploy.md
```diff
@@ -0,0 +1,247 @@
...
+
+To automate the trigger of Falco deployment when upstream aka origin repo creates an event.
+We will then deploy the benchmarking workfload for the project, in this case Falco.
+
+### Goals
+
+- Trigger GitHub Action workflow in green-reviews-tooling repo when Falco needs to be tested
+- Ask Falco team to implement the trigger
+- Deploy correct version of Falco in GitHub Action using Flux
+- Test the deployment via the Falco trigger
```
rossf7, line 55: @dipankardas011 As suggested by @leonardpahlke could you add a goal for documentation? 

To describe the implementation and how to configure new projects.

docs/proposals/proposal-trigger-and-deploy.md
```diff
@@ -0,0 +1,247 @@
...
+CNCF projects will be given a GitHub webhook they can call to trigger the
+green reviews pipeline. The webhook will accept 3 inputs and trigger a GitHub
+Actions workflow.
+
+Inputs are
+
+- `cncf_project`: **required** Project to be deployed e.g. `falco`
+- `cncf_project_sub`: **optional** Subcomponent if project has multiple variants
+they wish to test e.g. `modern-ebpf`
+- `version`: **required** Version of project to be tested e.g. `0.37.0`
```
rossf7, line 154: @nikimanoledaki @leonardpahlke @AntonioDiTuri @locomundo At this point the design has a dependency on the other proposals. 

As any inputs needed for the benchmark and report steps will also need to be added here. 

It would be good to define these early so we can present a complete solution to the Falco team.

docs/proposals/proposal-trigger-and-deploy.md
```diff
@@ -0,0 +1,247 @@
...
+
+Project resources that should always be deployed in the cluster are stored in
+the current location in the tooling repo e.g. `/clusters/projects/falco/`
+and are reconciled by source-controller.
+
+When the pipeline executes it will look for yaml files in the projects dir.
+If there is a yaml file matching the `cncf_project` input its contents will be
+applied using kubectl. The same applies for the `cncf_project_sub` input. 
+
+The `version` param will need to be injected into the files to ensure the
```
rossf7, line 183: @dipankardas011 Did I understand correctly your idea is to use kustomize to inject the version param? Could you explain a bit more?

docs/proposals/proposal-trigger-and-deploy.md
dipankardas011: I have added the commit hope it helps

docs/proposals/proposal-trigger-and-deploy.md
dipankardas011: I have added the commit hope it helps where we might move towards kustomize but still at initial phase lets make it very simple and use kubernetes raw manifests with repeations and later we can easily optimise, if we need!

rossf7: @dipankardas011 Good call, lets keep it simple for applying the manifests

docs/proposals/proposal-trigger-and-deploy.md
```diff
@@ -0,0 +1,286 @@
...
+defining a trigger mechanism, involving the Falco team in the implementation, 
+deploying Falco using Flux, and testing the deployment process. In future the 
+pipeline will support more CNCF projects as they are onboarded.
+
+The proposal also includes considerations for a phased implementation of the 
+automation, starting with manual triggering followed by automation via a webhook.
+
+The high level architecture is shown in this diagram.
+
+![wg green reviews workflow vision](./files/green-reviews-wg-workflow-vision.webp)
```
rossf7, line 51: @dipankardas011 I added the diagram as is from the blog post. I think it adds context and my bad for not including any diagrams in the previous version! 

I think an updated diagram focused on trigger & deploy would still be useful if you have time. WDYT?

rossf7 (commented):
Proposal is updated to include subscribing to new Falco releases.

The lightest touch way I can see to do that is via the releases atom feed. @leonardpahlke thank you this was great feedback on the previous version!

@nikimanoledaki PTAL when you have time

docs/proposals/proposal-trigger-and-deploy.md
dipankardas011: Sure

docs/proposals/proposal-trigger-and-deploy.md
```diff
@@ -0,0 +1,286 @@
...
+    ├── falco.yaml
+    ├── kmod.yaml
+    └── modern-ebpf.yaml
+```
+
+The pipeline will use a GitHub secret that has a kubeconfig to access the
+green reviews cluster. The deploy step in the pipeline will wait for the newly
+created flux resources to be reconciled before proceeding to the run step.
+
+We will have a node to deploy the project and another to run the benchmarks
```
SFulpius, line 238: Do you mean this? Otherwise I don't understand why node is mentioned here.
```suggestion
We will have a job to deploy the project and another to run the benchmarks
```

docs/proposals/proposal-trigger-and-deploy.md
```diff
@@ -0,0 +1,286 @@
...
+clusters/projects/${project_name}
+
+# and for each cncf_project (different configurations)
+clusters/projects/${project_name}/${configuration_name}
+```
+
+e.g. `/clusters/projects/falco/`
+and are reconciled by source-controller.
+
+When the pipeline executes it will look for yaml files in the projects dir.
```
SFulpius, line 217: What is the use of these yaml files? The project is already deployed using flux.

docs/proposals/proposal-trigger-and-deploy.md
```diff
@@ -0,0 +1,286 @@
...
+automation, starting with manual triggering followed by automation via a webhook.
+
+The high level architecture is shown in this diagram.
+
+![wg green reviews workflow vision](./files/green-reviews-wg-workflow-vision.webp)
+
+## Motivation
+
+To automate the trigger of Falco deployment when upstream aka origin repo 
+creates an event. We will then deploy the benchmarking workload for the
```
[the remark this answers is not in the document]
rossf7: Yes, its whenever a new release is created

docs/proposals/proposal-trigger-and-deploy.md
```diff
@@ -0,0 +1,286 @@
...
+
+```bash
+curl -X POST \
+     -H "Accept: application/vnd.github.v3+json" \
+     -H "Authorization: token $GITHUB_PAT" \
+     https://api.github.com/repos/cncf-tags/green-reviews-tooling/actions/workflows/pipeline.yaml/dispatches \
+     -d '{"ref":"main", "inputs": {"cncf_project": "falco", "cncf_project_sub": "modern-ebpf","version":"0.37.0"}}'
+```
+
+The projects will also be able to call this webhook using a fine grained access
```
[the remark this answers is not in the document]
rossf7: Yes, its the maintainers of the CNCF project and Falco is the first project. I'll make that clearer.

docs/proposals/proposal-trigger-and-deploy.md
```diff
@@ -0,0 +1,286 @@
...
+reviews pipeline.
+
+#### Project maintainer deploys their project so it can be measured
+
+Participating CNCF projects will deploy their project using a gitops approach
+with flux. This is described in more detail in the design details section.
+
+#### Project maintainer triggers pipeline to measure a release
+
+Calling the webhook will trigger the pipeline. The Report stage will provide
```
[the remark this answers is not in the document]
rossf7: Yes, release automation is automatic. The story name is outdated here. I'll change it to test a new benchmark.

docs/proposals/proposal-trigger-and-deploy.md
rossf7: For the first implementation all components including Falco and the benchmarks are deployed 24/7 by Flux.

We want to change that so the flux resources only exist for the lifetime of the workflow run. This will allow us to test more configurations of Falco and later other CNCF projects using the same cluster nodes.

Thanks that's great feedback that this isn't clear. I'll rework this section.

docs/proposals/proposal-trigger-and-deploy.md
rossf7: The cluster currently has 2 worker nodes. One node is used to deploy falco and the other node has the rest of the components including the benchmarks, prometheus, grafana etc.

This is to mitigate noisy neighbour type problems affecting the energy measurements of falco. Same here I'll rework this section to be clearer.

rossf7 (commented):
@SFulpius Thank you, that's great feedback. There are a few sections that need to be clearer. I'll address that in the next iteration.

docs/proposals/files/trigger-deploy-pipeline.svg
dipankardas011, line 1: Added the design of the pipeline flow

> [!NOTE]
> benchmarking part is left gray as this will be defined in later proposals

dipankardas011: was thinking about should we use github repo variables for storing the state and if any update is found it will update the github repo variable 

Refer: https://docs.github.com/en/rest/actions/variables

rossf7: @dipankardas011 Couple of minor things on diagram

- There should be arrows from Create Resources and Delete Resources and the Equinix cluster
- Typo on "Benchmark job" 🤓 

> was thinking about should we use github repo variables for storing the state and if any update is found it will update the github repo variable

Yes! I think that's a better way to manage the state. Its only needed by the workflow and updating the YAML file can cause merge conflicts. 

I updated the `Subscribing to Releases` section to reflect this

dipankardas011:
@rossf7 should we add the github variable part as well or we should test once we almost got the implementation and then we can optimize the variable thing before finishing the issue?

rossf7:
> should we add the github variable part as well 

@dipankardas011 I added a note about using variables in the `Subscribing to Releases` section. Is that OK or do you think more detail is needed?

dipankardas011:
> > should we add the github variable part as well 
> 
> @dipankardas011 I added a note about using variables in the `Subscribing to Releases` section. Is that OK or do you think more detail is needed?
> 

Yes then it's okay, thanks for adding that

docs/proposals/files/trigger-deploy-pipeline.svg
leonardpahlke, line 1: @dipankardas011 👍 the diagram looks a lot better, very nice.

docs/proposals/files/trigger-deploy.yml
```diff
@@ -0,0 +1,30 @@
...
+        description: 'Version'
+        required: true
+        default: '0.37.0'
+
+jobs:
+  echo-inputs:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
```
leonardpahlke, line 25: there is a `actions/checkout@v4` version available

docs/proposals/files/trigger-deploy.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    - name: Echo Inputs
```
leonardpahlke, line 27: Echo Inputs? is this for debugging? What are the inputs we echo - a project reference?

docs/proposals/proposal-001-trigger-and-deploy.md
leonardpahlke, line 1: Browsing over the doc - looks good! Usually, in projects, we and merge the proposal before we open an implementation PR ((or multiple PRs). So proposal and implementation is split. Just smth I noticed, feel free to shape the process who you all think works best @rossf7 @nikimanoledaki @AntonioDiTuri

rossf7: I've removed the example workflow. Agree, we should keep proposals and implementation separate.

docs/proposals/files/trigger-deploy.yml
rossf7: The workflow was intended to be an example to help with reviewing the PR.

However I don't think its adding much and is a distraction so I've removed it

docs/proposals/proposal-001-trigger-and-deploy.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Trigger
+
+The green reviews pipeline will be triggered by sending a [workflow_dispatch](https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event)
+event via the GitHub REST API.
+
+Inputs are
+
+- `cncf_project`: **required** Project to be deployed e.g. `falco`
+- `cncf_project_sub`: **optional** Subcomponent if project has multiple variants
```
ADiTuri, line 151: Is this a list? Shall we add it in the docs? Would it be helpful to have an example with multiple project_sub?

ADiTuri:
LGTM, I added just a minor comment, great job!

docs/proposals/proposal-001-trigger-and-deploy.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Drawbacks (Optional)
+
+<!--
+What other approaches did you consider, and why did you rule them out? These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+
+## Alternatives
```
ADiTuri, line 231: Maybe here as alternative we could link the discussion that we had once with @dipankardas011 about the pull vs push mechanism. From what I got we now decided to go with a pull mechanism - as checking if a new version is deployed - but we also evaluated the push mechanism - the project actively triggering the action, we discarded this and we could maybe document why. What do you think?

docs/proposals/proposal-001-trigger-and-deploy.md
```diff
@@ -0,0 +1,248 @@
...
+
+<!--
+List criteria which would allow progression from one maturity level to another.
+eg. What needs to have been accomplished/demonstrated to move from Alpha to Beta.
+
+If applicable, what is the milestone marker which will allow deprecation of the
+replaced capability?
+-->
+
+## Drawbacks (Optional)
```
ADiTuri, line 223: One possible drawback of the pull approach might be that we need to check every 30min. WDYT?

docs/proposals/proposal-001-trigger-and-deploy.md
rossf7: @ADiTuri Thank you that needed clarifying.

Its an optional list. Falco have 3 drivers they wish to use but so far we only deploy a single driver. 
I added this example to make it clearer. Does that help?

docs/proposals/proposal-001-trigger-and-deploy.md
rossf7: Push based is still supported. The pipeline can be triggered by calling the GitHub REST API with the token we will provide.

I added pull based after feedback from @leonardpahlke that just push based requires projects to change their CI/CD to add the trigger. 

@ADiTuri @dipankardas011 I added a note on this. WDYT?

docs/proposals/proposal-001-trigger-and-deploy.md
rossf7: @ADiTuri Good point I added this as a drawback but it can be mitigated by triggering via the REST API.

docs/proposals/proposal-001-trigger-and-deploy.md
dipankardas011: yes lets add this

and yes why to bother the cncf project maintainers for our pipeline trigger the easy thing would be to have every 30 minutes sync for fetuching latest repo and then trigger. I think this is much more attainable and easier to maintain and secure

docs/proposals/proposal-001-trigger-and-deploy.md
dipankardas011: interesting
but if the cron job is only fetching the latest release versions from github api then it is okay

docs/proposals/proposal-001-trigger-and-deploy.md
AntonioDiTuri: Yes I also think pull based works better, just wanted to have documented that we had this discussion and we took a decision between the two alternatives. Great like this, feel free to resolve the conversation

AntonioDiTuri (approved):
LGTM, seems like we only have one pending minor discussion, let's wait maximum a week and we can close this

docs/proposals/proposal-001-trigger-and-deploy.md
AntonioDiTuri: Maybe even once a day? I feel the frequency of the releases won't be so high right? Anyway we can state in the proposal that this parameter can be configurable!

docs/proposals/proposal-trigger-and-deploy.md
AntonioDiTuri: As we agreed during the meeting this will be a live document - let's edit this (if needed) after proposal 2 and 3 are finalized

docs/proposals/proposal-001-trigger-and-deploy.md
rossf7: @ADiTuri Yes, releases won't happen often but I think daily could be too infrequent 
I've increased this to 60 minutes as suggested by @dipankardas011 

We can also easily change this during implementation if we feel hourly is too often.

docs/proposals/proposal-001-trigger-and-deploy.md
```diff
@@ -0,0 +1,233 @@
...
+to only allow a single execution of the pipeline at any one time.
+
+The separate nodes are a best practice to prevent other components affecting
+the energy measurements.
+
+### Cleanup
+
+On completion of the pipeline run whether it was successful or failed the flux
+resources will be deleted via kubectl. The pipeline will wait for the flux
+resources to be deleted before exiting.
```
nikimanoledaki, line 209: This depends on which CLI was used to create the resources (see above for pros/cons of flux vs kubectl). If the resources were created using the `flux` CLI, it would be best to delete them using the `flux` CLI as well rather than `kubectl` since those resources would be controlled by Flux resources.

nikimanoledaki (commented):
Really great work and progress on this proposal, well done and thank you @dipankardas011 @rossf7! 🎉 🌟

docs/proposals/proposal-001-trigger-and-deploy.md
```diff
@@ -0,0 +1,233 @@
...
+    ├── kmod.yaml
+    └── modern-ebpf.yaml
+```
+
+The pipeline will use a GitHub secret that has a kubeconfig to access the
+green reviews cluster. The deploy step in the pipeline will wait for the newly
+created flux resources to be reconciled before proceeding to the run step.
+
+We will have a node to deploy Falco and another to run the benchmarks
+so we will use [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency)
```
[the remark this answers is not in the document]
rossf7: > Hm, should they be running on the same Node?

True, for Falco they need to run on the same node. 

I think we still need the concurrency limit. I'll reword this section.

docs/proposals/proposal-001-trigger-and-deploy.md
```diff
@@ -0,0 +1,233 @@
...
+
+The CNCF projects will be given a GitHub fine grained access token limited to
+the `green-reviews-tooling` repo. This token will have
+
+- Read access to code and metadata
+- Read write access to actions
+
+### Deploy
+
+Flux is used to deploy the CNCF project. Projects are able to use either
```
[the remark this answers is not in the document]
rossf7: @nikimanoledaki Right now the proposal is to use kubectl. I like that from a simplicity POV and because it makes deletion easier via `kubectl delete -f`. 

However the pros you make for using the Flux CLI make sense to me.

In an attempt to reduce cognitive load I would be in favour of aligning how we deploy the projects in prop 1 and the benchmarks in prop 2.

If we're agreed on that I can update the proposal so the pipeline dispatcher workflow calls a falco workflow that deploys using the Flux CLI?

cc @ADiTuri @dipankardas011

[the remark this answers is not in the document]
dipankardas011: okay kubectl 👍🏼

docs/proposals/proposal-001-trigger-and-deploy.md
dipankardas011: > How do we trigger the delete? How about adding an action_type input to the workflow with deploy or uninstall?

> In the workflow we can have multiple jobs that run depending on the input. In the uninstall job we can add the flux delete command.

@rossf7 was not able to understand this part
even if we have running benchmark test it will trigger deletion once the benchmarking is completed. so may be we can start with this or something which you are refering to?

rossf7: > Let's add a note in the proposal that we should run a step that deletes the project that was deployed.

@nikimanoledaki I updated the Cleanup section to have more detail on the deletion. PTAL

> was not able to understand this part

@dipankardas011 Sorry for causing confusion. This was if we were going to delete with the flux CLI. However we decided to use kubectl so this doesn't apply.

locomundo: I was just thinking... how does the Trigger and Deploy pipeline know when the benchmarks are finished, in order to clean up? In general, how is the communication between the two?

Something related: when deploying the CNCF project (I suppose with `kubectl apply`) we should follow with a `kubectl wait --for condition`, so that the benchmarking only starts after everything is up and running, I suppose?

And for the benchmarks, per the last meeting discussion (maybe I should take this to the other PR):
- If the benchmark tests have a start and an end, there should be some kind of signal that this pipeline knows when to clean up (or perhaps it just knows to wait until the Benchark pipeline(s) are done?)
- If it's a test which runs on an interval of time, like the `stress-ng` for 15 minutes, who determines when to start and stop it? I suppose that is part of the Benchmarking step/pipeline to decide? But this is related to the unit R of the formula.

Maybe I'm complicating things and all of this is already clear, I'd like to hear everybody's thoughts..

rossf7: > how does the Trigger and Deploy pipeline know when the benchmarks are finished, in order to clean up? In general, how is the communication between the two?

@locomundo For a successful pipeline run this will be after the benchmarks have run and the metrics are saved to long term storage. I think this should be covered in proposal 3 but I added a note on that to the Cleanup section.

> Something related: when deploying the CNCF project (I suppose with kubectl apply) we should follow with a kubectl wait --for condition, so that the benchmarking only starts after everything is up and running, I suppose?

Yes we will use `kubectl apply -f` to apply the manifests but good point we'll also need to use `kubectl wait` before starting the benchmarking to ensure flux has reconciled the resources. I added this to the Deploy section.

> If it's a test which runs on an interval of time, like the stress-ng for 15 minutes, who determines when to start and stop it? I suppose that is part of the Benchmarking step/pipeline to decide? But this is related to the unit R of the formula.

I think the benchmarking step will need to control this but I think its best to include that in the proposal 2 PR.

dipankardas011 (commented):
LGTM

nikimanoledaki (approved):
LGTM2! 🚀

rossf7:
First proposal down 🚀 

Thank you @dipankardas011 for all your hard work on it and to all the reviewers! 💚

PR #92, opened 2024-05-06
Update PUE for equinix data centers according to 2023 report

raymundovr:
#### What type of PR is this?
kind/enhancement

Equinix has recently published a new value for the PUE of their data centers from their 2023 report. The current value is not up to date with it.

None


rossf7 (approved):
LGTM

PR #102, opened 2024-06-04
fix: Updates to proposal 1

rossf7:
#### What type of PR is this?

kind/documentation

Update proposal 1 to 

- Make releases check weekly instead of hourly. As we feel hourly is too often.
- Use releases endpoint instead of Atom feed
- Use JSON not YAML for projects metadata to avoid need for Go YAML library

Also adds missing link to atom feed.

None

cc @dipankardas011

### DIFF
--- a/docs/proposals/proposal-001-trigger-and-deploy.md
+++ b/docs/proposals/proposal-001-trigger-and-deploy.md
@@ -129,8 +129,12 @@ projects:
       - kmod
 ```
 
-A scheduled GitHub Action will run every hour and check the Atom feed of
-each project for new releases. To manage the state a GitHub [repository variable](https://docs.github.com/en/actions/learn-github-actions/variables)
+A scheduled GitHub Action will run weekly and check the Atom feed of
+each project for new releases.
+
+e.g. https://github.com/falcosecurity/falco/releases.atom
+
+To manage the state a GitHub [repository variable](https://docs.github.com/en/actions/learn-github-actions/variables)
 per CNCF project is used to store the latest release version.
 
 If a new release is detected the action will trigger the pipeline for the new
--- a/docs/proposals/proposal-001-trigger-and-deploy.md
+++ b/docs/proposals/proposal-001-trigger-and-deploy.md
@@ -57,7 +57,7 @@ multiple configurations of Falco and more CNCF projects as they are onboarded.
 
 ### Goals
 
-- Trigger the pipeline when a new release of a project happens
+- Trigger the pipeline once a day for the latest version of the project
 - Allow additional runs of the pipeline by calling a GitHub webhook
 - Deploy the new version of the project using flux
 - Delete the resources at the end of the pipeline run
@@ -75,7 +75,7 @@ team
 
 ## Proposal
 
-We will watch for new releases of the project via the GitHub REST API
+Once a day we will get the latest release of the project via the GitHub REST API
 e.g. https://api.github.com/repos/falcosecurity/falco/releases/latest
 
 Our automation will call the GitHub REST API to trigger the pipeline.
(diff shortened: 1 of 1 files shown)


dipankardas011:
@rossf7 as you are adding the atom release here may be we can change the purpose of this pr itself if its okay?
and make the necessary changes to accommodate for the jq one
WDYT?

rossf7:
> as you are adding the atom release here may be we can change the purpose of this pr itself if its okay?

@dipankardas011 Yes sure. Done

dipankardas011 (approved):
LGTM

dipankardas011 (approved):
Lgtm

rossf7 (commented):
Updating the spec after pairing session with @dipankardas011 to simplify the design.

Changes are

- We no longer use a github var to store the latest release
- Instead we trigger the pipeline once a day for the latest version of Falco

Not needing to maintain state simplifies and allows us to use bash for the workflow 🎉 

@nikimanoledaki @AntonioDiTuri PTAL 🙏

docs/proposals/proposal-001-trigger-and-deploy.md
```diff
@@ -57,7 +57,7 @@ multiple configurations of Falco and more CNCF projects as they are onboarded.
 
 ### Goals
 
-- Trigger the pipeline when a new release of a project happens
+- Trigger the pipeline once a day for the latest version of the project
```
dipankardas011, line 60: @rossf7 what about we trigger the workflow for the benchmarking in a specific time period where it is cheap and lower load on grid like at night may be where the k8s cluster is located
WDYT?

dipankardas011 (approved):
LGTM

docs/proposals/proposal-001-trigger-and-deploy.md
rossf7: @dipankardas011 Thanks good to consider this. I updated the spec to say 08:00 UTC.

This is because for renewable energy I think the best time to run it is in daylight hours when demand is low. As there should be solar energy available then. 

Although the cluster is in France so the average carbon intensity is low because the grid has a lot of nuclear.

docs/proposals/proposal-001-trigger-and-deploy.md
```diff
@@ -116,33 +116,38 @@ removed. In future we could create nodes on demand and delete on completion.
...
-      - ebpf
-      - modern-ebpf
-      - kmod
+```json
+{
+    "projects": [
+        {
+            "name": "falco",
+            "organization": "falcosecurity",
+            "sub_components": [
```
nikimanoledaki, line 128: We discussed in the meeting that we could rename this to `configurations` rather than `sub_components`

rossf7: @nikimanoledaki Yes, I think configurations is a much better name.

Proposal updated to reflect this and the workflow input is renamed from `cncf_project_sub` to `config`.

docs/proposals/proposal-001-trigger-and-deploy.md
```diff
@@ -116,39 +116,44 @@ removed. In future we could create nodes on demand and delete on completion.
...
+A JSON file of CNCF projects will be stored in the green-reviews-tooling repo.
+Optionally multiple configurations of a CNCF project can be deployed e.g.
+
+```json
+{
+    "projects": [
+        {
+            "name": "falco",
+            "organization": "falcosecurity",
+            "configurations": [
```
nikimanoledaki, line 128:
```suggestion
            "config": [
```

nikimanoledaki (approved):
LGTM 👍 

CI is failing with:
```
Run bash project-trigger.sh
GH_TOKEN not set
Error: Process completed with exit code 20.
```

Should the trigger script be running on PRs?

rossf7:
@nikimanoledaki @dipankardas011 

Thank you! I'll create an implementation issue for the renaming.

> Should the trigger script be running on PRs?

No I think it should only run if there is a change to scripts dir.

Proposed a fix in #109

Issue #100, opened 2024-06-04
[ACTION] Modify flux resources for Falco to support pipeline


rossf7:
# Task Description

To support the deploy step changes in https://github.com/cncf-tags/green-reviews-tooling/issues/99 we need to modify the flux resources we use to deploy Falco.

We need to sync with the Falco team on the changes.

## Current State

- Falco is deployed 24/7 using flux with the resources in [clusters/projects/falco](https://github.com/cncf-tags/green-reviews-tooling/tree/main/clusters/projects/falco)
- This references the https://github.com/falcosecurity/cncf-green-review-testing repo maintained by the Falco team
- Resources for all 3 falco drivers are created but only `modern-ebpf` pods are scheduled due to node labels

## Desired State

- Falco is only deployed during a run of the green review pipeline
- `config` input is used so only resources for a single Falco driver are created
- `version` input is used to template the container image that is deployed
- New node selector used for all drivers `node-role.kubernetes.io/benchmark: "true"`

## Tasks

- [ ] Delete flux resources from green-reviews-tooling so falco is not deployed 24/7 - [code](https://github.com/cncf-tags/green-reviews-tooling/tree/main/clusters/projects/falco)
- [ ] Refactor flux resources in falcosecurity/cncf-green-review-testing to support new inputs - [spec](https://github.com/cncf-tags/green-reviews-tooling/blob/main/docs/proposals/proposal-001-trigger-and-deploy.md#trigger)

rossf7:
## Status

@nikimanoledaki @AntonioDiTuri Posting an update here as part of my handover since I'll be off till 15th July. Lets discuss in our sync.

Changes to the Falco manifests are proposed in this [discussion](https://github.com/falcosecurity/cncf-green-review-testing/discussions/17#discussioncomment-9936561).

If Falco team are happy with the changes then once merged we need to update each kustomization resource to use the new paths. 

e.g. https://github.com/cncf-tags/green-reviews-tooling/blob/main/projects/falco/ebpf.yaml#L19 

```
path: ./kustomize/falco-driver/ebpf
```

This PR has cluster changes to support the deploy step https://github.com/cncf-tags/green-reviews-tooling/pull/111

PR #104, opened 2024-06-08
feature: add release fetcher and dispatcher

### LINKED ISSUE #98: [ACTION] Add release trigger
# Task Description

Implement trigger for a green review when there is a new release of a participating CNCF project. Falco is the first project.

Design is described in proposal 1.
https://github.com/cncf-tags/green-reviews-tooling/blob/main/docs/proposals/proposal-001-trigger-and-deploy.md#subscribing-to-releases

## Tasks

- [ ] Create GitHub Actions workflow for release trigger that runs weekly - [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule)
- [ ] Add `clusters.yaml` with falco and sub components for drivers (`modern-ebpf`, `kmod`, `ebpf`) - [spec](https://github.com/cncf-tags/green-reviews-tooling/blob/main/docs/proposals/proposal-001-trigger-and-deploy.md#subscribing-to-releases)
- [ ] Check GitHub REST API for new releases - [spec](https://github.com/cncf-tags/green-reviews-tooling/blob/main/docs/proposals/proposal-001-trigger-and-deploy.md#subscribing-to-releases)
- [ ] Send workflow dispatch event to trigger pipeline workflow - [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch)
- [ ] Save latest release per project to GitHub repository variable - [docs](https://docs.github.com/en/actions/learn-github-actions/variables)

dipankardas011:
#### What type of PR is this?
kind/feature

TBD

Fixes #98

### DIFF
--- a/.github/workflows/project-trigger.yml
+++ b/.github/workflows/project-trigger.yml
@@ -1,0 +1,37 @@
+name: Reconciles Project Versions and triggers them
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  get-latest-version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '^1.22.3'
+
+      - name: build the binary
+        working-directory: cmd/github-project-update-dispatch
+        run: |
+          go get -d
+          go build -v -o ../../proj-trigger .
+
+      - name: trigger updated projects
+        env:
+          GH_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
+          LOG_LEVEL: DEBUG # yet to use this
+        run: |
+          ./proj-trigger -c projects/projects.json
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 infrastructure/equinix-metal/.terraform/
 infrastructure/equinix-metal/terraform.tfvars
 infrastructure/equinix-metal/.terraform.lock.hcl
+website/public/
+website/resources/
+website/node_modules/
+.idea
--- a/cmd/github-project-update-dispatch/flags.go
+++ b/cmd/github-project-update-dispatch/flags.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"os"
+)
+
+func initProjects() (resp *Projects, err error) {
+	var locProjectMetaFile string
+
+	flag.StringVar(&locProjectMetaFile, "c", "projects/projects.json", "Full path of the metadata file for projects")
+	flag.Parse()
+
+	_b, err := os.ReadFile(locProjectMetaFile)
+	if err != nil {
+		return
+	}
+
+	err = json.Unmarshal(_b, &resp)
+	return
+}
--- a/cmd/github-project-update-dispatch/github_api.go
+++ b/cmd/github-project-update-dispatch/github_api.go
@@ -1,0 +1,65 @@
+package main
+
+import (
+	"fmt"
+	"net/url"
+	"os"
+	"strings"
+)
+
+func NewGHRepoVarStorage() (*GithubRepository, error) {
+	v, ok := os.LookupEnv(EnvGithubPatVarName)
+	if !ok || len(v) == 0 {
+		return nil, fmt.Errorf("the environment variable for github pat is missing")
+	}
+	return &GithubRepository{githubToken: v}, nil
+}
+
+func generateVariableName(projName string) string {
+	return strings.ToLower(projName + "_version")
+}
+
+func gitHubRepoEndpoint(
+	org, repo string,
+	operation githubAPIEndpointType,
+	variableName, workflowFileName string) (result string, err error) {
+
+	baseUrl := "https://api.github.com"
+	urlPath := []string{"repos", org, repo, "actions"}
+
+	switch operation {
+	case GHRepoVariableEndpoint:
+		urlPath = append(urlPath, "variables", variableName)
+	case GHWorkflowDispatchEndpoint:
+		urlPath = append(urlPath, "workflows", workflowFileName, "dispatches")
+	case GHLatestRelease:
+		urlPath = append(urlPath, "releases", "latest")
+	}
+
+	result, err = url.JoinPath(baseUrl, urlPath...)
+	return
+}
+
+func (obj *GithubRepository) genURLAndHeaders(
+	endpointType githubAPIEndpointType,
+	variableName string, workflowFileName string,
+) (*string, map[string]string, error) {
+
+	url, err := gitHubRepoEndpoint(
+		GHRepoOrganizationName,
+		GHRepoName,
+		endpointType,
+		variableName,
+		workflowFileName,
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return &url,
+		map[string]string{
+			"Accept":               "application/vnd.github+json",
+			"Authorization":        "Bearer " + obj.githubToken,
+			"X-GitHub-Api-Version": "2022-11-28",
+		}, nil
+}
--- a/cmd/github-project-update-dispatch/github_api_repo_vars.go
+++ b/cmd/github-project-update-dispatch/github_api_repo_vars.go
@@ -1,0 +1,125 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"strings"
+	"time"
+)
+
+func (obj *GithubRepository) UpdateRepoVariable(projName, newVersion string) error {
+	variableName := generateVariableName(projName)
+
+	url, header, err := obj.genURLAndHeaders(GHRepoVariableEndpoint, variableName, "")
+	if err != nil {
+		return err
+	}
+
+	newVariableData := struct {
+		Name string `json:"name"`
+		Val  string `json:"value"`
+	}{
+		Name: strings.ToUpper(variableName),
+		Val:  newVersion,
+	}
+
+	var _newVariableData bytes.Buffer
+
+	if err := json.NewEncoder(&_newVariableData).Encode(newVariableData); err != nil {
+		return fmt.Errorf("failed to serialize the body: %v", err)
+	}
+
+	resp, err := NewHTTPClient(
+		http.MethodPatch,
+		*url,
+		time.Minute,
+		&_newVariableData,
+		header,
+	)
+	if err != nil {
+		return err
+	}
+
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusNoContent {
+		return fmt.Errorf("status code was not 204, got: %v", resp.StatusCode)
+	}
+
+	return nil
+}
+
+func (obj *GithubRepository) ReadRepoVariable(projName string) (*string, error) {
+	variableName := generateVariableName(projName)
+	url, header, err := obj.genURLAndHeaders(GHRepoVariableEndpoint, variableName, "")
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := NewHTTPClient(
+		http.MethodGet,
+		*url,
+		time.Minute,
+		nil,
+		header,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("status code was not 200, got: %v", resp.StatusCode)
+	}
+
+	var variableData struct {
+		VariableValue string `json:"value"`
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&variableData); err != nil {
+		return nil, fmt.Errorf("failed deserialize response body: %v", err)
+	}
+
+	return &variableData.VariableValue, nil
+}
+
+func fetchLatestRelease(org, proj string) (*string, error) {
+
+	url, err := gitHubRepoEndpoint(
+		GHRepoOrganizationName,
+		GHRepoName,
+		GHLatestRelease,
+		"", "",
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := NewHTTPClient(
+		http.MethodGet,
+		url,
+		time.Minute, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("status code was not 200, got: %v", resp.StatusCode)
+	}
+
+	var release struct {
+		TagName string `json:"tag_name"`
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&release); err != nil {
+		return nil, fmt.Errorf("failed deserialize response body: %v", err)
+	}
+
+	slog.Info("Latest Release", "Proj", proj, "Org", org, "Ver", release.TagName)
+
+	return &release.TagName, nil
+}
--- a/cmd/github-project-update-dispatch/github_api_repo_workflows.go
+++ b/cmd/github-project-update-dispatch/github_api_repo_workflows.go
@@ -1,0 +1,75 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"time"
+)
+
+func (obj *GithubRepository) DispatchCall(projData Project, version string) error {
+	projToWorkflowTable := map[string]string{
+		"falco": "falco.yaml",
+	}
+
+	workflowFileName, found := projToWorkflowTable[projData.Name]
+	if !found {
+		return fmt.Errorf("failed to find the workflow file related to project: %s", projData.Name+"/"+projData.Organization)
+	}
+
+	url, header, err := obj.genURLAndHeaders(GHWorkflowDispatchEndpoint, "", workflowFileName)
+	if err != nil {
+		return err
+	}
+
+	type WorkflowData struct {
+		Ref    string `json:"ref"`
+		Inputs struct {
+			CncfProject    string `json:"cncf_project"`
+			Version        string `json:"version"`
+			CncfProjectSub string `json:"cncf_project_sub"`
+		} `json:"inputs"`
+	}
+
+	for i := range projData.SubComponents {
+		workflowData := WorkflowData{}
+		workflowData.Inputs.CncfProject = projData.Name
+		workflowData.Ref = "main" // TODO: need to figure it out
+		workflowData.Inputs.Version = version
+		workflowData.Inputs.CncfProjectSub = projData.SubComponents[i]
+
+		var _newWorkflowData bytes.Buffer
+
+		if err := json.NewEncoder(&_newWorkflowData).Encode(workflowData); err != nil {
+			return fmt.Errorf("failed to serialize the body: %v", err)
+		}
+
+		resp, err := NewHTTPClient(
+			http.MethodPost,
+			*url,
+			time.Minute,
+			&_newWorkflowData,
+			header,
+		)
+		if err != nil {
+			return err
+		}
+
+		defer resp.Body.Close()
+		if resp.StatusCode != http.StatusNoContent {
+			return fmt.Errorf(
+				"failed in workflowDispatch, status code was not 204, got: %v",
+				resp.StatusCode)
+		}
+
+		slog.Info("Successfully workflow_dispatch",
+			"Proj", projData.Name,
+			"Org", projData.Organization,
+			"Ver", version,
+			"Subcomponent", projData.SubComponents[i])
+	}
+
+	return nil
+}
--- a/cmd/github-project-update-dispatch/github_api_repo_workflows.go
+++ b/cmd/github-project-update-dispatch/github_api_repo_workflows.go
@@ -10,16 +10,8 @@ import (
 )
 
 func (obj *GithubRepository) DispatchCall(projData Project, version string) error {
-	projToWorkflowTable := map[string]string{
-		"falco": "falco.yaml",
-	}
 
-	workflowFileName, found := projToWorkflowTable[projData.Name]
-	if !found {
-		return fmt.Errorf("failed to find the workflow file related to project: %s", projData.Name+"/"+projData.Organization)
-	}
-
-	url, header, err := obj.genURLAndHeaders(GHWorkflowDispatchEndpoint, "", workflowFileName)
+	url, header, err := obj.genURLAndHeaders(GHWorkflowDispatchEndpoint, "", "deploy.yaml")
 	if err != nil {
 		return err
 	}
--- a/cmd/github-project-update-dispatch/go.mod
+++ b/cmd/github-project-update-dispatch/go.mod
@@ -1,0 +1,10 @@
+module github.com/cncf-tags/green-reviews-tooling/cmd/github-project-update-dispatch
+
+go 1.22.3
+
+require gotest.tools v2.2.0+incompatible
+
+require (
+	github.com/google/go-cmp v0.6.0 // indirect
+	github.com/pkg/errors v0.9.1 // indirect
+)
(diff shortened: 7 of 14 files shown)


dipankardas011:
## Questions
- How are we planning to store what all projects are supported? (is it some environment variable or a file as a variable in repo?)
- found a hard limit for github variable is 500 only

## plans

will using ${{ vars.<> }} to get eh variable once we get the version will do a curl request to the following repo to get the current latest version if we need to update it will update that via github rest api and once done need to trigger the following pipeline for respective projects

the dispatch we need to figure it out

rossf7:
> How are we planning to store what all projects are supported? (is it some environment variable or a file as a variable in repo?)

I think we should use a file. I included an example file in the proposal in the `subscribing to releases` section but it may need more data such as the GitHub repo URL .
https://github.com/cncf-tags/green-reviews-tooling/blob/main/docs/proposals/proposal-001-trigger-and-deploy.md#subscribing-to-releases

For the location I think it should be `/projects/project.yaml`? As underneath the projects dir we will have the per project manifests. e.g. `/projects/falco/ebpf.yaml` WDYT?

> found a hard limit for github variable is 500 only

Good to check this. However if we use 1 variable per project I don't think we'll ever hit this limit. Or if we do hit it our automation will need to be more sophisticated and later we can switch to something else for managing state.

> how to update the github repo variable

Can you use the REST API?

dipankardas011:
> Can you use the REST API?
Forgot to update was able to work with that just that we need to use the same curl command to retrive the variable

planning to use python script to as much as possible no dependency install

dipankardas011:
> For the location I think it should be /projects/project.yaml? As underneath the projects dir we will have the per project manifests. e.g. /projects/falco/ebpf.yaml WDYT?

SGTM, I also want to specifiy the organization of the project as well

dipankardas011:
@rossf7 Can you create a PAT token for only this repo so that we can see if it works or not?

dipankardas011:
Here are the variables I am looking for

secrets.REPO_ACCESS_TOKEN -> for read and write of the repo variables
vars.FALCO_VERSION -> some random string as value
vars.KEPLER_VERSION -> some random string as value

rossf7:
> planning to use python script to as much as possible no dependency install

@dipankardas011  Reasoning makes sense but as discussed we need agreement that python is the higher level language we want to use. I'll discuss with the other leads.

> Can you create a PAT token for only this repo so that we can see if it works or not?

I don't have permission to create tokens. Can you test with your fork for now? Once we're aligned on the python question we can request the token.

dipankardas011:
> I don't have permission to create tokens. Can you test with your fork for now? Once we're aligned on the python question we can request the token.

Tested it works 😁

dipankardas011:
Just thought that python would be better as it's much more maintainable than writting complicated bash scripts😉

dipankardas011:
Still would consider others review as I am not that great in python 🙂

.github/workflows/falco.yml
```diff
@@ -0,0 +1,25 @@
...
+        required: true
+        type: string
+
+jobs:
+  TBD:
+    runs-on: ubuntu-latest
+    steps:
+      - name: check the triggered call
+        run: |
+          echo "Version: ${{ inputs.image_tag }}"
```
rossf7, line 25: What is this workflow for?

We will need to trigger the deploy workflow being added in #105 passing the 3 inputs.

Since we're using the REST API to fetch the releases I think we should also use it to trigger.

projects/project.json
```diff
@@ -0,0 +1,17 @@
...
+            "name": "falco",
+            "organization": "falcosecurity",
+            "sub_components": [
+                "ebpf",
+                "modern-ebpf",
+                "kmod"
+            ]
+        },
+        {
+            "name": "kepler",
```
rossf7, line 13: Kepler should not be here. Falco is the only project we are measuring.

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,284 @@
...
+		return nil, fmt.Errorf("failed deserialize response body: %v", err)
+	}
+
+	slog.Info("Latest Release", "Proj", proj, "Org", org, "Ver", release.TagName)
+
+	return toPtr[string](release.TagName), nil
+}
+
+type GithubVariable struct {
+	gh_token     string
```
rossf7, line 96:
```suggestion
	githubToken     string
```

Nit but lets use camel case for all the vars

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,284 @@
...
+	fullRepoName string
+}
+
+func NewGithubVariable() (*GithubVariable, error) {
+	const EnvGithubPat string = "GH_TOKEN"
```
rossf7, line 101: Lets move this to the top of the file with the other constants

rossf7 (commented):
@dipankardas011 Thank you for your efforts here and being flexible on using Go for this rather than Python.

As discussed ideally we would use bash but the parsing of the releases endpoint response and error handling make this tricky so I think using Go makes sense.

Design wise my main question is how the deploy step will be triggered? I think we should use the REST API as its just another endpoint to call.

I'll reply separately on the repo config changes.

rossf7:
> Here are the variables I am looking for

```
secrets.REPO_ACCESS_TOKEN: for read and write of the repo variables
vars.FALCO_VERSION: some random string as value
```

For `REPO_ACCESS_TOKEN` we have an existing secret called `FLUX_GITHUB_TOKEN`. I think we should rename this to `GITHUB_TOKEN` and we can use it for both tasks.

https://github.com/cncf-tags/green-reviews-tooling/blob/main/.github/workflows/tofu.yaml#L23

`FALCO_VERSION` yes we will need to add this variable. I checked the endpoint and unfortunately there is no "upsert" support so this will be a setup task when we onboard new projects which we'll need to document.

dipankardas011:
> Design wise my main question is how the deploy step will be triggered? I think we should use the REST API as its just another endpoint to call.

Yes I thought about it but then another problem cam arise as we are directly calling a different workflow we can't know for sure which parent called the child workflow. so thought to give workflow_call method as I have used previously and it works nice
not sure if it answers your question :)

.github/workflows/falco.yml
dipankardas011: actually my purpose for the version as the only input here is as we are sending the version which we want to install rest of the checks which all component are present in the projects/project.json might be possible from the child job to take out and use it

still if you want to add more info we can.

and the updatedProjects are stored in /tmp/updatedProjects.json and it is used to set the environment variable for particular proj so that we can add a if statement for each job and it will be triggered only if its true

projects/project.json
dipankardas011: yeah will remove it 😄

scripts/repo-variable-writer.go
dipankardas011: sure

dipankardas011:
>  I checked the endpoint and unfortunately there is no "upsert" support so this will be a setup task when we onboard new projects which we'll need to document.

📓 Keep a note

dipankardas011:
> For REPO_ACCESS_TOKEN we have an existing secret called FLUX_GITHUB_TOKEN. I think we should rename this to GITHUB_TOKEN and we can use it for both tasks.

Does `FLUX_GITHUB_TOKEN` contains repo:variable:write perms?

rossf7:
> Yes I thought about it but then another problem cam arise as we are directly calling a different workflow we can't know for sure which parent called the child workflow. so thought to give workflow_call method as I have used previously and it  works nice

@dipankardas011 This should be implemented via a workflow_dispatch event as was agreed in the proposal.
https://github.com/cncf-tags/green-reviews-tooling/blob/main/docs/proposals/proposal-001-trigger-and-deploy.md#trigger

The 3 inputs mean we know the project, version and which falco driver to deploy.

> Does FLUX_GITHUB_TOKEN contains repo:variable:write perms?
> I would say lets keep it specific to perms and role for that PAT token so that we can know what is it used for
> also general aka all in one token doesn't seems correct as each token should have very limited scope

Good point. The current token doesn't have permission and better to scope it to the needed permissions.

So we will need to request a new token.

dipankardas011:
> > Yes I thought about it but then another problem cam arise as we are directly calling a different workflow we can't know for sure which parent called the child workflow. so thought to give workflow_call method as I have used previously and it  works nice
> 
> @dipankardas011 This should be implemented via a workflow_dispatch event as was agreed in the proposal.
> https://github.com/cncf-tags/green-reviews-tooling/blob/main/docs/proposals/proposal-001-trigger-and-deploy.md#trigger
> 
> The 3 inputs mean we know the project, version and which falco driver to deploy.
> 
> > Does FLUX_GITHUB_TOKEN contains repo:variable:write perms?
> > I would say lets keep it specific to perms and role for that PAT token so that we can know what is it used for
> > also general aka all in one token doesn't seems correct as each token should have very limited scope
> 
> Good point. The current token doesn't have permission and better to scope it to the needed permissions.
> 
> So we will need to request a new token. 
> 
> 
> 
> 
> 

Okay so you meant instead of calling like
Uses: ./.GitHub......
We should use the GitHub api call instead to call the workflow dispatch ?

May be it's not that different still if we do it we might be better of calling the workflows from the go script itself

.github/workflows/get-latest-releases.yml
```diff
@@ -0,0 +1,57 @@
...
+          fi
+
+
+  falco-proj:
+    uses: ./.github/workflows/falco.yml
+    needs: ["get-latest-version"]
+    if: ${{ needs.check-updates.outputs.is-falco-updated }} == 'true'
+    secrets: inherit
+    with:
+      version: ${{ vars.FALCO_VERSION }} # need to check if it gets the updated value
```
dipankardas011, line 54: So this thing we should change with api calls as well?

dipankardas011:
@rossf7 should we add context as well for each fmt.Errorf to get which component failed ?

scripts/repo-variable-writer.go
leonardpahlke, line 1: repo variable writer, i dont know what that means.

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+
+	req, err := http.NewRequest(method, url, body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create newRequest, method: %s, url: %s, Reason: %v", method, url, err)
+	}
+
+	for k, v := range headers {
+		req.Header.Add(k, v)
+	}
+
```
leonardpahlke, line 50: we may want to add a debug step here.

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+	type WorkflowData struct {
+		Ref    string `json:"ref"`
+		Inputs struct {
+			CncfProject    string `json:"cncf_project"`
+			Version        string `json:"version"`
+			CncfProjectSub string `json:"cncf_project_sub"`
+		} `json:"inputs"`
+	}
+
+	for _, subComponent := range projData.SubComponents {
```
leonardpahlke, line 255: usually we want to use the index and avoid copy things around.


```go
for i := range projData.SubComponents {
// ...
workflowData.Inputs.CncfProjectSub = projData.SubComponents[i]
```

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+const (
+	LocProjMeta         string = "../projects/project.json"
+	EnvGithubPatVarName string = "GH_TOKEN"
+)
+
+func toPtr[T any](v T) *T {
+	return &v
+}
```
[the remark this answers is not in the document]
leonardpahlke: i expect that this code only reads and does not write anything right? it does not make any changes to a repository?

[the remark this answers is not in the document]
leonardpahlke: it would be good to create some kind of README that explains how to use the project, where it is used and what it does & why it exists

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+
+func loadProjectMetadata(locProj string) (resp *Projects, err error) {
+	_b, err := os.ReadFile(locProj)
+	if err != nil {
+		return
+	}
+
+	err = json.Unmarshal(_b, &resp)
+	return
+}
```
leonardpahlke, line 66: this looks like it should not be a function

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+
+func fetchLatestRelease(org, proj string) (*string, error) {
+	url := fmt.Sprintf(
+		"https://api.github.com/repos/%s/%s/releases/latest", org, proj)
```
leonardpahlke, line 70: Maybe there is a way to make this "fetch release" which also has the capability to fetch the latest release. Giving functions more depth and avoid shallow functions

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+
+	slog.Info("Latest Release", "Proj", proj, "Org", org, "Ver", release.TagName)
+
+	return toPtr[string](release.TagName), nil
+}
+
+type GithubVariable struct {
+	githubToken  string
+	fullRepoName string
+}
```
leonardpahlke, line 102: GithubVariable is a confusing name. Its more a GithubRepository

leonardpahlke: or smth. even just `GH` would be a bit better IMO

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+		return nil, fmt.Errorf("failed deserialize response body: %v", err)
+	}
+
+	return toPtr[string](variableData.VariableValue), nil
+}
+
+func (obj *GithubVariable) UpdatedProjectDispatcher(projData Project, version string) error {
+
+	projToWorkflowTable := map[string]string{
+		"falco": "falco.yaml",
```
leonardpahlke, line 233: this probably should not get hardcoded

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+func main() {
+
+	projects, err := loadProjectMetadata(LocProjMeta)
+	if err != nil {
+		slog.Error("Failed: unable to read the projects Metadata", "Reason", err)
+		os.Exit(1)
+	}
+	slog.Info("Got projects metadata", "Projects", projects)
+
+	storage, err := NewGithubVariable()
```
leonardpahlke, line 305: storage?

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+			}
+
+			storedRelease, err := storage.ReadRepoVariable(proj.Name)
+			if err != nil {
+				errChan <- err
+				return
+			}
+
+			if strings.Compare(*storedRelease, *latestRelease) != 0 {
+				slog.Info("Release Versions", "Proj", proj.Name, "Org", proj.Organization, "Latest", *latestRelease, "Current", *storedRelease)
```
leonardpahlke, line 337: maybe more a debug log. we need to make sure we do not clutter our stack.

leonardpahlke: this log does not tell us much. its debug information.

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+		return fmt.Errorf("failed to serialize the body: %v", err)
+	}
+
+	resp, err := handleHTTPCall(
+		http.MethodPatch,
+		*url,
+		time.Minute,
+		&_newVariableData,
+		header,
+	)
```
leonardpahlke, line 182: why do we make changes to the repository?

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+	slog.Info("Got projects metadata", "Projects", projects)
+
+	storage, err := NewGithubVariable()
+	if err != nil {
+		slog.Error("Failed: initialize storage", "Reason", err)
+		os.Exit(1)
+	}
+
+	wg := new(sync.WaitGroup)
+	wg.Add(len(projects.Projects))
```
leonardpahlke, line 312: 👍 good

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+		if err != nil {
+			hadFailures = true
+			slog.Error("Failed", "Reason", err)
+		}
+	}
+	if hadFailures {
+		os.Exit(1)
+	}
+
+	slog.Info("Finished execution")
```
leonardpahlke, line 372: this does not provide any information. we can either remove this log or add some kind of info to it

leonardpahlke:
wem need to add some quality checks to the repo. this can be done in another PR. golint gosec etc.

scripts/repo-variable-writer.go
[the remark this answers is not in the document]
dipankardas011: @leonardpahlke Thanks for your review

Will try to update my nvim conf for gofmt.

not sure if we should create a go pkg @rossf7

scripts/repo-variable-writer.go
```diff
@@ -0,0 +1,373 @@
...
+	Organization  string   `json:"organization"`
+	SubComponents []string `json:"sub_components"`
+}
+
+type Projects struct {
+	Projects []Project `json:"projects"`
+}
+
+const (
+	LocProjMeta         string = "../projects/project.json"
```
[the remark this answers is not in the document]
dipankardas011: okay

scripts/repo-variable-writer.go
dipankardas011: sure will change the name once the structure is finialized (initial name ;) )

scripts/repo-variable-writer.go
dipankardas011: You mean slog.Debug()??

scripts/repo-variable-writer.go
dipankardas011: I'm considering github repo variables as a persistent storage, thats why this name

scripts/repo-variable-writer.go
dipankardas011: It doesn't change the repo but change the specific repository variables to store the new versions

scripts/repo-variable-writer.go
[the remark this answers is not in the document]
dipankardas011: not able to configure to use gofumpt, neither goland nor nvim

scripts/repo-variable-writer.go
dipankardas011: not sure what you meant my shallow function in this context

scripts/repo-variable-writer.go
dipankardas011: yes currently wait for ross to look at this and I can then update the logic here

like do we have different workflow files or a single workflow file like in #105

scripts/repo-variable-writer.go
dipankardas011: I think we shouldn't attach debug log here as it will contain the Github PAT TOken

scripts/repo-variable-writer.go
rossf7: > yes currently wait for ross to look at this and I can then update the logic here

> like do we have different workflow files or a single workflow file like in https://github.com/cncf-tags/green-reviews-tooling/pull/105

@dipankardas011 There will be a single workflow file. The project name is passed to the workflow in the `cncf_project` input.

leonardpahlke:
@dipankardas011 i will propose some edits in a PR and open it towards your repository.

leonardpahlke:
Alright, I was looking at the code. And also the plan we worked on so far. 

My understanding, looking at the code.
1. load a project manifest json from the green-reviews repo which contains the information about all projects that we like to deploy. the file reference can be passed as flag. `--repo-manifest https://github.com/cncf-tags/green-reviews-tooling/blob/main/repo-manifests.json`

```json
{
  "falco": {
    "repo": "https://github.com/falcosecurity/falco"
  }
}
```

2. loop over all repositories and ? 

We need the helm charts or some other commands that we can execute to deploy Falco. Falco's helm charts seem to be located here https://github.com/falcosecurity/charts/tree/master/charts/falco if there is a new Falco release, we would deploy the helm chart. So the information we need is a bit more.

```json
{
  "falco": {
    "v0.38": { // config used for version v0.38 or newer unless a higher version is specified. in case the deployment changes by release
      "repo": "https://github.com/falcosecurity/falco",
      "deployment": [
        "helm repo add falcosecurity https://falcosecurity.github.io/charts",
        "helm repo update",
        "helm install falco falcosecurity/falco --create-namespace --namespace falco"
      ],
      "verify": [],          // to check if the deployment succeeded
      "collectdata": [],     // idk how this is done. maybe automatically
      "destroy": []          // delete any deployed resources
    }
  }
}

```

You are right, @dipankardas011 we need some kind of storage to keep track of the latest release measured by project. A repository variable could do the trick `"falco:v0.38,xyz:v0.1"`.

Right now, if i read the current plan right ([plan](https://github.com/cncf-tags/green-reviews-tooling/blob/main/docs/proposals/proposal-001-trigger-and-deploy.md#trigger)), we have one pipeline per project (falcosecurity/falco) right now and use the workflow-dispatch action to trigger our deployment pipeline. Perhaps its easier to trigger the deployment pipeline once a day. We do not need to measure the minute a release is cut. This would make it easier. We can always manually trigger the pipeline if we wish to do so or increase the cadance. I would try to reduce complexity as much and involve more and more folks in a process only if that is really necessary.

leonardpahlke:
This could be done in a bash script.

rossf7:
@leonardpahlke Thank you for reviewing. Adding my thoughts on these design changes.

Step 1:  yes, the release trigger is to check if there is a new release of Falco. It should call the deploy workflow being added in #105

This was added so we could subscribe to releases without projects needing to call our deploy workflow from their CI/CD. Although they can do that if they wish.

Step 2: The Falco team wished to deploy using Kustomize and created this repo.
https://github.com/falcosecurity/cncf-green-review-testing

There are 3 Falco drivers they wish to test. For #100 my plan is to create an overlay per driver using kustomize.

> This could be done in a bash script.

Yes, when I talked with Niki and Antonio our preference would be to use bash and only use go if absolutely necessary.

I like your idea of always triggering the pipeline once a day for the latest release. This would be useful for testing since Falco releases do not happen often and mean we don't need to store state.

I think simplifying the design so we can use bash is the best way to move forward.

rossf7:
@leonardpahlke @dipankardas011 I've been thinking more on how we could simplify the design. As I think I made it more complex than was necessary :(

Instead of the projects.json I think we could use inputs and default to the falco values.

Dipankar, I think your previous suggestion to use workflow_call instead of workflow_dispatch to call the deploy workflow is better and mean a PAT is not needed to trigger the deploy.

WDYT?

dipankardas011:
> Dipankar, I think your previous suggestion to use workflow_call instead of workflow_dispatch to call the deploy workflow is better and mean a PAT is not needed to trigger the deploy.

Sol 1> For that we can make the go script which writes a file in json and parse and set environment variables

somthing like this (previous impl)

```yaml
jobs:
  get-latest-version:
    runs-on: ubuntu-latest
    outputs:
      is-falco-updated: ${{ steps.check-updates.outputs.FALCO }}
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-go@v5
        with:
          go-version: '^1.22.3'

      - name: build the binary
        # env:
        #   GH_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
        working-directory: cmd
        run: |
          go get -d
          go build -v -o ../proj-trigger .

      - name: trigger updated projects
        # env:
        #   GH_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
        working-directory: cmd
        run: |
          go get -d
          go build -v -o ../proj-trigger .

      - name: Dispatch
        id: check-updates
        env:
          file: /tmp/updates.json # generated by go program
        run: |
          if jq -e --arg val "falco" '.proj_names | index($val) != null' "$file" > /dev/null; then
            echo "Element '$value' found in the proj_names array."
            echo "FALCO=true" >> $GITHUB_OUTPUT
          else
            echo "Element 'falco' not found in the proj_names array."
            echo "FALCO=false" >> $GITHUB_OUTPUT
          fi


  falco-proj:
    uses: .github/workflows/falco.yml
    needs: ["get-latest-version"]
    if: ${{ needs.check-updates.outputs.is-falco-updated }} == 'true'
    secrets: inherit
    with:
      version: ${{ vars.FALCO_VERSION }} # need to check if it gets the updated value
  ```
  
Sol 2> We can do this way to have a seperate bash script or another option which is go will write the bash script which contains these echo "FALCO=false" >> $GITHUB_OUTPUT
 and we will just perform source and make a command like
 set_updated_projects_envs() in the bash
 
 ```bash
 go run ......
 # next step
 source ~/${some source} # written by go
 set_updated_projects_envs()
 ```
 
 by this the envs for these will be written I beloeve to the $GITHUB_OUTPUT
 
 other than that not sure if we should move entirely to bash scripts (not very comfortable[would need help]) and remove go script. Still decision on you both

dipankardas011:
Also I have made to many files than necessary (actually was trying to undertand leo's inputs). may be that looks too much complicated

dipankardas011:
@leonardpahlke I have closed this PR and created a new one as its will potentially cause review problems so now Using Bash script with droped Idea of project version to github variables and instead to call it everyday to perform for all projects irrespective of if there is or not there any new version
#105

Issue #7, opened 2023-11-22
[Tracking/Request] Create GitHub PAT to use with Flux


nikimanoledaki:
Unblocks https://github.com/cncf-tags/green-reviews-tooling/issues/5

We want to use Flux as part of our GitOps approach to deploying the applications in the cluster that we are setting up for the sustainability assessments.

We would like to request a GitHub PAT from the `cncf-tags` org, please 

The Flux docs specify the requirements for the GitHub PAT for an Organization [here](https://fluxcd.io/flux/installation/bootstrap/github/#github-organization) - the docs recommend creating a bot account:
> If you want to bootstrap Flux for a repository owned by an GitHub organization, it is recommended to create a dedicated user for Flux under your organization.

> Generate a GitHub PAT for the Flux user that can create repositories by checking all permissions under repo.

> If you want to use an existing repository, the Flux user must have admin permissions for that repository.

We can use a [fine-grained PAT](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#fine-grained-personal-access-tokens) since they are more secure than classic PATs.

RobertKielty:
Hi @nikimanoledaki and @leonardpahlke

I've created a fine-grained token, wg-green-review-flux that grants the following access to the cncf-tags/green-reviews-tooling repo:

**Repository permissions**
 * Read access to metadata
 * Read and Write access to administration and code
 
I've shared the token with @leonardpahlke

leonardpahlke:
Thanks @RobertKielty !

nikimanoledaki:
Thank you @RobertKielty! 

We'll be able to close this once we test the PAT as part of https://github.com/cncf-tags/green-reviews-tooling/issues/5 👍

AntonioDiTuri:
I tested this and it works! I think we can close this issue! Thanks everybody

nikimanoledaki:
Wonderful, thank you for testing, @AntonioDiTuri! 🎉

rossf7:
Reopening as for #99 we have a new workflow that we'd like to trigger via the REST API using [workflow_dispatch](https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event)

https://github.com/cncf-tags/green-reviews-tooling/blob/main/.github/workflows/benchmark-pipeline.yaml

For that we need a new fine grained token with the permissions below.

> "Actions" repository permissions (write)

[...]

PR #106, opened 2024-06-19
feat: Trigger other Benchmark workflows

### LINKED ISSUE #98: [ACTION] Add release trigger

dipankardas011:
#### What type of PR is this?
kind/feature

- Improvises on #104 

- Fixes #98

### DIFF
--- a/.github/workflows/project-trigger.yml
+++ b/.github/workflows/project-trigger.yml
@@ -2,7 +2,7 @@ name: Reconciles Project Versions and triggers them
 
 on:
   schedule:
-    - cron: '0 0 * * 0'
+    - cron: '0 8 * * *'
   push:
     branches: [main]
   pull_request:
--- a/.github/workflows/project-trigger.yml
+++ b/.github/workflows/project-trigger.yml
@@ -1,0 +1,30 @@
+name: Reconciles Project Versions and triggers them
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  get-latest-version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: |
+          sudo apt update -y
+          sudo apt install jq -y
+
+      - name: trigger the workflows
+        working-directory: scripts
+        env:
+          GH_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
+        run: |
+          bash project-trigger.sh
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 infrastructure/equinix-metal/.terraform/
 infrastructure/equinix-metal/terraform.tfvars
 infrastructure/equinix-metal/.terraform.lock.hcl
+website/public/
+website/resources/
+website/node_modules/
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,4 @@
 infrastructure/equinix-metal/.terraform/
 infrastructure/equinix-metal/terraform.tfvars
 infrastructure/equinix-metal/.terraform.lock.hcl
-website/public/
-website/resources/
-website/node_modules/
 .idea
--- a/scripts/project-trigger.sh
+++ b/scripts/project-trigger.sh
@@ -18,28 +18,28 @@ jq -c '.projects[]' "$json_file" | while read -r project; do
     proj_organization=$(echo "$project" | jq -r '.organization')
     sub_components=$(echo "$project" | jq -r '.sub_components')
 
-    echo "Project Name: $proj_name"
-    echo "Organization: $proj_organization"
-    echo "SubComponents: $sub_components"
+    echo "[DBG] Project Name: $proj_name"
+    echo "[DBG] Organization: $proj_organization"
+    echo "[DBG] SubComponents: $sub_components"
 
     release_url="https://api.github.com/repos/${proj_organization}/${proj_name}/releases/latest"
     
     
-    response=$(curl -fsSL -X GET $release_url)
+    response=$(curl --fail-with-body -sSL -X GET $release_url)
     status_code=$?
     if [ $status_code -ne 0 ]; then
-        echo "Error fetching latest release for ${proj_name} from ${release_url}. Status code: $status_code"
+        echo "[ERR] fetching latest release for ${proj_name} from ${release_url}. Status code: $status_code"
         continue
     fi
 
     latest_proj_version=$(echo "$response" | jq -r '.tag_name')
     if [ -z "$latest_proj_version" ]; then
-        echo "Error: Could not find the latest version for ${proj_name}"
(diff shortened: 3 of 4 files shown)


dipankardas011:
Only place left for testing this script is workflow_dispatch with 2 types of workflows which can have sub_components or not have it

also need to think of do we send one sub_component at a time or all at once like a json array or a string with each having its individual workflow_call?

.gitignore
```diff
@@ -1,3 +1,7 @@
 infrastructure/equinix-metal/.terraform/
 infrastructure/equinix-metal/terraform.tfvars
-infrastructure/equinix-metal/.terraform.lock.hcl
\ No newline at end of file
+infrastructure/equinix-metal/.terraform.lock.hcl
+website/public/
+website/resources/
+website/node_modules/
+.idea
```
rossf7, line 7: Needs a rebase. These lines were removed.

scripts/project-trigger.sh
```diff
@@ -0,0 +1,61 @@
...
+
+json_file="../projects/projects.json"
+gh_token=$GH_TOKEN
+git_ref="main"
+workflow_organization_name="cncf-tags"
+workflow_project_name="green-reviews-tooling"
+workflow_dispatcher_file_name="benchmark-pipeline.yaml"
+
+if [ -z "$gh_token" ]; then
+    echo "[ERR] GH_TOKEN not set"
```
rossf7, line 12:
```suggestion
    echo "GH_TOKEN not set"
```

Some output is important but lets keep it as simple as possible. 

We don't need these prefixes so pls remove them from all the echo statements. If its an error the echo text should explain that.

scripts/project-trigger.sh
```diff
@@ -0,0 +1,61 @@
...
+
+    latest_proj_version=$(echo "$response" | jq -r '.tag_name')
+    if [ -z "$latest_proj_version" ]; then
+        echo "[ERR] Could not find the latest version for ${proj_name}"
+        continue
+    fi
+
+    echo "[DBG] Version: $latest_proj_version"
+
+    # TODO(dipankar): need to make the cncf_project_sub to be each individiual call for each sub-component
```
rossf7, line 44: Yes, this still needs to be implemented.

.github/workflows/project-trigger.yml
```diff
@@ -0,0 +1,30 @@
+name: Reconciles Project Versions and triggers them
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'
```
rossf7, line 5:
```suggestion
    - cron: '0 0 * * *'
```

We changed design to run daily rather than weekly.

scripts/project-trigger.sh
```diff
@@ -0,0 +1,61 @@
...
+
+    status_code=$?
+    if [ $status_code -ne 0 ]; then
+        echo "[ERR] dispatching workflow for ${proj_name}. Status code: $status_code"
+        echo "[ERR] Response: $workflow_dispatch"
+        continue
+    fi
+
+    echo "[INF] workflow_call event [proj: $proj_name]=> $workflow_dispatch"
+    echo "-----------------------------"
```
rossf7, line 60:
```suggestion
    echo "workflow_call event [proj: $proj_name]=> $workflow_dispatch"
```

rossf7 (changes requested):
@dipankardas011 Thank you the approach here using bash works a lot better.

Some changes still needed. Especially the output to the console which can be much simpler.

scripts/project-trigger.sh
rossf7: > also need to think of do we send one sub_component at a time or all at once like a json array or a string with each having its individual workflow_call?

The benchmark-pipeline action needs to be called once per subcomponent if they exist.

Its an array in the projects.json so can you loop through each component? 

If there are no sub components you could pass an empty string.

dipankardas011:
@rossf7 Should I add color for error logs?

rossf7 (commented):
> @rossf7 Should I add color for error and success logs?

@dipankardas011 No that isn't necessary and makes the code more complex.

Pls revert the changes in 129d1105b09e97ee048d1aca110c549e77a3f691

Better to focus on making sure the text in the echo statements is clear. I suggested a couple of edits and then this is good to go

scripts/project-trigger.sh
```diff
@@ -0,0 +1,95 @@
...
+    latest_proj_version=$(echo "$response" | jq -r '.tag_name')
+    if [ -z "$latest_proj_version" ]; then
+        echo "could not find the latest version for ${proj_name}"
+        continue
+    fi
+
+    echo "latest version: $latest_proj_version"
+
+    if [ "$sub_components" == "null" ]; then
+        echo "No sub-components for $proj_name. Performing certain task..."
```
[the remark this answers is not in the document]
dipankardas011: Done

scripts/project-trigger.sh
```diff
@@ -0,0 +1,95 @@
...
+        status_code=$?
+        if [ $status_code -ne 0 ]; then
+            echo "dispatching workflow for [proj: ${proj_name}, ver: $latest_proj_version] Status code: $status_code"
+            echo "curl response: $workflow_dispatch"
+            continue
+        fi
+
+        echo "workflow_call event [proj: $proj_name, ver: $latest_proj_version]=> OK"
+    else
+        echo "$proj_name has sub-components. Performing tasks for each sub-component..."
```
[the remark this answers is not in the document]
dipankardas011: Done

rossf7 (approved):
LGTM 💚

dipankardas011:
> LGTM 💚

But the env var for the gh_pat token?

rossf7:
> But the env var for the gh_pat token?

@dipankardas011 We need to open a service desk ticket to request a new token. I'll take care but will need some help from the other leads.

rossf7:
@dipankardas011 Token is added as a secret and working well here.
https://github.com/cncf-tags/green-reviews-tooling/issues/7#issuecomment-2189406546

Nice job! 🚀 

There is an issue with the deploy step itself. Will debug that later

PR #107, opened 2024-06-24
docs: Create kubeconfig for pipeline

rossf7:
#### What type of PR is this?

kind/documentation

Follow up to #105 

Docs to generate a kubeconfig for the pipeline so it can be added as a GitHub secret

Cluster role has view access + full access to Flux custom resources which is needed by the [pipeline](https://github.com/cncf-tags/green-reviews-tooling/blob/main/.github/workflows/benchmark-pipeline.yaml)


dipankardas011 (approved):
LGTM

the bash script to accept options is nice

nikimanoledaki (approved):

nikimanoledaki:
awesome 🌟 thank you for adding a separate kubeconfig for the pipeline!

PR #108, opened 2024-06-26
fix: Quote kubeconfig and add sleep to deploy step

rossf7:
#### What type of PR is this?

kind/bug

- Adds quoting to echo statement when creating kubeconfig from secret
- Adds a sleep for 10 seconds before kubectl wait command to allow flux resources to be created.

Follow up to #99 

cc @dipankardas011

### DIFF
--- a/.github/workflows/benchmark-pipeline.yaml
+++ b/.github/workflows/benchmark-pipeline.yaml
@@ -24,7 +24,7 @@ jobs:
         with:
           version: v1.30.2
         id: install
-      - run: mkdir ~/.kube && echo ${{ secrets.KUBECONFIG }} > ~/.kube/config
+      - run: mkdir ~/.kube && echo "${{ secrets.KUBECONFIG }}" > ~/.kube/config
       - name: Select the manifest
         run: |
           MANIFEST=projects/${{ inputs.cncf_project }}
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark-pipeline.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
         with:
           version: v1.30.2
         id: install
-      - run: mkdir ~/.kube && echo ${{ secrets.KUBECONFIG }} > ~/.kube/config
+      - run: mkdir ~/.kube && echo "${{ secrets.KUBECONFIG }}" > ~/.kube/config
```
rossf7, line 27: kind error was a red herring due to `kind: Config` in the kubeconfig.

The echo just needs to be quoted to preserve whitespace.

.github/workflows/benchmark-pipeline.yaml
```diff
@@ -51,6 +51,8 @@ jobs:
         run: |
           kubectl apply -f manifest.yaml
 
+          sleep 10
```
rossf7, line 54: Without the sleep the wait command fails with `error: no matching resources found` 

Not the most graceful of fixes 🔨  but it works 😄

PR #109, opened 2024-06-28
fix: Only trigger on script changes and align naming

rossf7:
#### What type of PR is this?

kind/bug

- Only run project trigger workflow for changes to scripts dir.
- Tidy up workflow naming for consistency

None

### DIFF
--- a/.github/workflows/benchmark-pipeline.yaml
+++ b/.github/workflows/benchmark-pipeline.yaml
@@ -1,4 +1,5 @@
-name: benchmark-pipeline
+name: Benchmark Pipeline
+
 on:
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmark-pipeline.yaml
```diff
@@ -1,4 +1,5 @@
-name: benchmark-pipeline
+name: Benchmark Pipeline
```
rossf7, line 1: We were using a different naming style in each workflow 🙈

.github/workflows/project-trigger.yml
```diff
@@ -1,12 +1,13 @@
...
   schedule:
     - cron: '0 8 * * *'
-  push:
-    branches: [main]
   pull_request:
-    branches: [main]
+    branches:
+    - main
+    paths:
+    - 'scripts/**'
```
rossf7, line 10: This is the actual fix so we only run if there is a change to the scripts dir.

nikimanoledaki (approved):
cchanges LGTM, thank you @rossf7! will rebase the Proposal 2 PR 👍

PR #122, opened 2024-09-11
Add benchmark_path field to projects data and benchmark-job to pipeline. (#118)

### LINKED ISSUE #118: [ACTION] Add `benchmark_path` field to projects.json and pipeline
Part of [Proposal-002 - Run the CNCF project benchmark tests as part of the automated pipeline
](https://github.com/cncf-tags/green-reviews-tooling/blob/main/docs/proposals/proposal-002-run.md)

# Task Description

This issue tracks implementing part of proposal 2. So the existing pipeline is extended to call the benchmark workflow defined in the `projects.json` metadata file.

#121 is related to create the benchmark workflow for Falco

## Tasks

- [ ] Add a new `benchmark_path` field in [projects.json](https://github.com/cncf-tags/green-reviews-tooling/blob/main/projects/projects.json).
- [ ] Update the [benchmark pipeline](https://github.com/cncf-tags/green-reviews-tooling/blob/main/.github/workflows/benchmark-pipeline.yaml) workflow to take `benchmark_path` as an input and call the benchmark workflow using the `uses` syntax

vallss:
#### What type of PR is this?
kind/enhancement

This PR adds the new field  `benchmark_path` to project.json and adds a new job to the benchmark pipeline that uses the specified path.

Fixes #118

### DIFF
--- a/.github/workflows/benchmark-pipeline.yaml
+++ b/.github/workflows/benchmark-pipeline.yaml
@@ -14,7 +14,7 @@ on:
         required: true
       benchmark_path:
         description: Path to the benchmark action
-        required: true
+        required: false # TODO: change to `true` when `"benchmark_path"` is specified in `projects/projects.json`
 
 concurrency:
   group: benchmark
--- a/.github/workflows/benchmark-pipeline.yaml
+++ b/.github/workflows/benchmark-pipeline.yaml
@@ -12,6 +12,9 @@ on:
       version:
         description: Version of project to be tested e.g. 0.37.0
         required: true
+      benchmark_path:
+        description: Path to the benchmark action
+        required: true
 
 concurrency:
   group: benchmark
@@ -58,6 +61,9 @@ jobs:
           --all \
           --for=condition=Ready \
           --namespace=benchmark
+
+  benchmark-job:
+    uses: ${{ inputs.benchmark_path }}
 
   delete:
     runs-on: ubuntu-22.04
--- a/projects/projects.json
+++ b/projects/projects.json
@@ -3,6 +3,7 @@
         {
             "name": "falco",
             "organization": "falcosecurity",
+            "benchmark_path": "",
             "configs": [
                 "ebpf",
                 "modern-ebpf",
--- a/scripts/project-trigger.sh
+++ b/scripts/project-trigger.sh
@@ -16,10 +16,12 @@ fi
 jq -c '.projects[]' "$json_file" | while read -r project; do
     proj_name=$(echo "$project" | jq -r '.name')
     proj_organization=$(echo "$project" | jq -r '.organization')
+    proj_benchmark_path=$(echo "$project" | jq -r '.benchmark_path')
     configs=$(echo "$project" | jq -r '.configs')
 
     echo "Project Name: $proj_name"
     echo "Organization: $proj_organization"
+    echo "Benchmark Path: $proj_benchmark_path"
     echo "Configs: $configs"
 
     release_url="https://api.github.com/repos/${proj_organization}/${proj_name}/releases/latest"


.github/workflows/benchmark-pipeline.yaml
```diff
@@ -12,6 +12,9 @@ on:
       version:
         description: Version of project to be tested e.g. 0.37.0
         required: true
+      benchmark_path:
+        description: Path to the benchmark action
+        required: true
```
rossf7, line 17: Makes sense for the input to be required but I think this will fail until we add the benchmark in the other ticket?

I think that's fine since we have the dependency between tickets but just wanted to mention it.

rossf7 (changes requested):
@vallss Thank you! This is looking good

I think we also need to extend the project trigger script to pass the benchmark path.
https://github.com/cncf-tags/green-reviews-tooling/blob/main/scripts/project-trigger.sh

Can you add that?

.github/workflows/benchmark-pipeline.yaml
vallss: Yes, I thought about it too and forgot to mention in the PR notes. 

I thought the field would be required but right now the job would obviously fail since a part of the logic is temporarily missing. Maybe I can add a comment? Or is it okay to keep track of this here?

nikimanoledaki: > Maybe I can add a comment?

+1 I would recommend changing this to false and adding a comment 👍 

```suggestion
        required: false # TODO: change to `true` when `"benchmark_path"` is specified in `projects/projects.json`
```

nikimanoledaki (approved):
LGTM with one change! 💚 Thank you for following up with the addition to `scripts/project-trigger.sh` from the meeting!

rossf7 (approved):
LGTM

rossf7:
@vallss Thank you for making the changes after the meeting.

After merging I see there is the error below https://github.com/cncf-tags/green-reviews-tooling/actions/runs/10847889481

```
Invalid workflow file
invalid value workflow reference: no version specified
```

Looks like its expecting an input like `actions/checkout@v2` with a version. Could you investigate this further?

.github/workflows/benchmark-pipeline.yaml
rossf7: @vallss @nikimanoledaki I'm sorry I missed the CI error after committing your suggestion Niki :( I was hoping to get this merged before a meeting but its Fri 13th and I should have been more cautious! 👻 🤦‍♂️ 

I was wondering if we could add an `if` condition for the input being empty but AIUI `uses` is a special case and can't evaluate an input. https://github.com/orgs/community/discussions/45342

I think it may be simpler to add an empty workflow for Falco and reference it via the filename in which case we don't need to provide a ref like `@main`

> ./.github/workflows/{filename} for reusable workflows in the same repository.

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_iduses

I think with our naming the filename would be `./.github/workflows/falco_benchmark.yml` @nikimanoledaki is that correct?

@vallss Do you have time to look at this?

Don't worry if not! I can also take a look at it next week.

vallss: @rossf7 @nikimanoledaki yes, no problem! I can take a look at it starting from Monday (I'm on a short holiday right now). 

This is related to this error, right?
```
Invalid workflow file: .github/workflows/benchmark-pipeline.yaml#L66
invalid value workflow reference: no version specified
```
Would this be correct when we set a path value?

rossf7: Thank you @vallss and yes this can definitely wait till you're back from your holiday!

> Would this be correct when we set a path value?

Yes, my reading of the [uses docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_iduses) is that if we specify a file path in the repo we don't need to specify an `@{ref}` and it will use the current commit.

However, I haven't done this before with GitHub Actions with a dynamic value for the workflow. So if you're able to try it out that would be great.

Likewise if you think there is a benefit to specifying an `@{ref}` we can use that approach instead.

PR #127, opened 2024-11-14
[Implementation] (Proposal 2)  Add Benchmark Pipeline 

dipankardas011:
#### What type of PR is this?
kind/feature

#### What this PR does / why we need it:
It will give us a alternative route to use kubernetes manifests for the benchmark jobs instead of reusing github workflow.

With this we are trying to use the manifest location in github or any public facing endpoint from where we can get them

#### Which issue(s) this PR fixes:

Addresses part of #121

#### Special notes for your reviewer (optional):

Created this for finding alternative route for the current proposal 2 impl

Refer: #124, https://github.com/cncf-tags/green-reviews-tooling/issues/121#issuecomment-2476378468

PR #130, opened 2024-11-26
fix: Don't run trigger on PR as secret not available

### LINKED ISSUE #129: [ACTION] Add approval process for running pipeline actions from forks
## Problem

The benchmarking pipeline uses 2 github secrets for the github access token and cluster kubeconfig.
By default pull requests from public forks can't access the secrets as its a potential security risk.

Allowing approved PRs to access these secrets will make them easier to review and improve the contributor experience.

## Possible Solutions

We should set up an approval process as described below.  Another common solution is for a reviewer to add a label to the PR that triggers the workflow.

https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks

rossf7:
#### What type of PR is this?

kind/bug

The project trigger action runs for some pull requests e.g. #127 but it uses a GH secret that is not accessible from forked repos. 

So it fails which is confusing when reviewing. This disables the action for pull requests only.

Fixes #129

### DIFF
--- a/.github/workflows/project-trigger.yml
+++ b/.github/workflows/project-trigger.yml
@@ -3,11 +3,6 @@ name: Project Trigger
 on:
   schedule:
     - cron: '0 8 * * *'
-  pull_request:
-    branches:
-    - main
-    paths:
-    - 'scripts/**'
   workflow_dispatch:
 
 concurrency:


locomundo (approved):
LGTM

PR #135, opened 2024-11-28
Move RBAC config to dir reconciled by Flux

nikimanoledaki:
#### What type of PR is this?

kind/enhancement

Moving RBAC config to `base/` so that it is reconciled by Flux.

None


rossf7 (approved):
LGTM

PR #139, opened 2024-11-28
fix: Add namespace to kubectl wait

rossf7:
#### What type of PR is this?

kind/bug

Fix for pipeline as without namespace `kubectl wait` fails

```
k --kubeconfig /tmp/pipeline-kubeconfig wait pod \
          --all \
          --for=condition=Ready
error: no matching resources found
```

```
k --kubeconfig /tmp/pipeline-kubeconfig wait pod \
          --all \
          --for=condition=Ready \
          --namespace=falco
pod/falco-event-generator-7f74b9cf77-s52fk condition met
pod/redis-7d65c855db-wtgk5 condition met
pod/stress-ng-85b4869f47-d72tj condition met
```

### DIFF
--- a/.github/workflows/benchmark-pipeline.yaml
+++ b/.github/workflows/benchmark-pipeline.yaml
@@ -103,7 +103,8 @@ jobs:
 
           kubectl wait pod \
           --all \
-          --for=condition=Ready
+          --for=condition=Ready \
+          --namespace=falco # TODO: Revert to "benchmark" this after merging https://github.com/falcosecurity/cncf-green-review-testing/pull/22
 
       - name: Wait for the benchmark job to complete
         run: |


.github/workflows/benchmark-pipeline.yaml
```diff
@@ -103,7 +103,8 @@ jobs:
 
           kubectl wait pod \
           --all \
-          --for=condition=Ready
+          --for=condition=Ready \
+          --namespace=falco # TODO: Revert to "benchmark" this after merging https://github.com/falcosecurity/cncf-green-review-testing/pull/22
```
dipankardas011, line 107: Instead of hardcoding we need to get the namespace name from the project.json

nikimanoledaki: We don't need to do that at the moment, IIUC we're using a generic `benchmark` namespace for all right now. I don't see us running 2 benchmark tests simultaneously in the future either

nikimanoledaki (approved):

PR #141, opened 2024-11-29
fix: Benchmark param name and add sleep prior to kubectl wait

rossf7:
#### What type of PR is this?

kind/bug

- Fixes incorrect param name in trigger script.
- Adds sleeps before wait in case there is a delay creating pods.
- Fixes namespace when waiting for falco pods.

### DIFF
--- a/.github/workflows/benchmark-pipeline.yaml
+++ b/.github/workflows/benchmark-pipeline.yaml
@@ -80,12 +80,12 @@ jobs:
         run: |
           kubectl apply -f manifest.yaml
 
-          sleep 10
+          sleep 20
 
           kubectl wait pod \
           --all \
           --for=condition=Ready \
-          --namespace=falco # TODO: Revert to "benchmark" this after merging https://github.com/falcosecurity/cncf-green-review-testing/pull/22
+          --namespace=benchmark
 
   benchmark-job:
     runs-on: ubuntu-24.04
@@ -100,6 +100,8 @@ jobs:
       - name: Run the benchmark job
         run: |
           kubectl apply -f ${{ inputs.benchmark_job_url }}
+          
+          sleep 20
 
           kubectl wait pod \
           --all \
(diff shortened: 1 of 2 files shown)


dipankardas011 (approved):

.github/workflows/benchmark-pipeline.yaml
```diff
@@ -80,12 +80,12 @@ jobs:
...
           kubectl apply -f manifest.yaml
 
-          sleep 10
+          sleep 20
 
           kubectl wait pod \
           --all \
           --for=condition=Ready \
-          --namespace=falco # TODO: Revert to "benchmark" this after merging https://github.com/falcosecurity/cncf-green-review-testing/pull/22
+          --namespace=benchmark
```
rossf7, line 88: Wait should match target namespace 
https://github.com/cncf-tags/green-reviews-tooling/blob/ae76ef9b13bb4e159e0ae56bae1f5f83f978e0a0/projects/falco/ebpf.yaml#L25

.github/workflows/benchmark-pipeline.yaml
```diff
@@ -80,12 +80,12 @@ jobs:
         run: |
           kubectl apply -f manifest.yaml
 
-          sleep 10
+          sleep 20
```
rossf7, line 83: This sleep wasn't long enough in some tests.

.github/workflows/benchmark-pipeline.yaml
```diff
@@ -100,6 +100,8 @@ jobs:
       - name: Run the benchmark job
         run: |
           kubectl apply -f ${{ inputs.benchmark_job_url }}
+          
+          sleep 20
```
rossf7, line 104: Added a sleep here too in case there is a delay creating the benchmark pods.

.github/workflows/benchmark-pipeline.yaml
nikimanoledaki: I see - it should be ok if we have falco deployed in namespace `benchmark` and the benchmark tests in namespace `falco`, temporarily. A bit confusing but we should be able to revert it soon. 🤞

nikimanoledaki (approved):

.github/workflows/benchmark-pipeline.yaml
rossf7: Yes it is a bit confusing but once your PR in the falco repo is merged we can tidy up and just use the benchmark namespace.

PR #146, opened 2024-12-11
Changing concurrency behavior to allow for queued pipelines

### LINKED ISSUE #142: [ACTION] - Investigate why not all falco configurations are benchmarked
With latest fixes the benchmark pipeline is triggered for each of the 3 falco configurations.

However only 2/3 pipeline runs are successful. The 3rd fails with an error that looks related to GitHub Actions concurrency setting 
https://github.com/cncf-tags/green-reviews-tooling/blob/45185e85d0c8e7d8a06009b306d1421ace9c749a/.github/workflows/benchmark-pipeline.yaml#L29-L30

```
Benchmark Pipeline
Canceling since a higher priority waiting request for 'benchmark' exists
```

e.g. https://github.com/cncf-tags/green-reviews-tooling/actions/runs/12155458645

SRF-Audio:
#### What type of PR is this?
kind/bug

This PR updates the concurrency behavior of the benchmark pipeline to address issues with job cancellations when multiple workflows are triggered simultaneously. Previously, only two pipelines could run at a time, and additional workflows were canceled due to the default concurrency settings.

Fixes #142 

Since this if for a workflow where I don't have the secrets/permissions to trigger, we'll need to verify that this change produces the desired behavior once the pipeline is triggered. According to the documentation, it should.

### DIFF
--- a/.github/workflows/benchmark-pipeline.yaml
+++ b/.github/workflows/benchmark-pipeline.yaml
@@ -28,6 +28,7 @@ on:
 
 concurrency:
   group: benchmark
+  cancel-in-progress: false
 
 jobs:
   print_summary:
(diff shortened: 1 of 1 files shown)


nikimanoledaki (approved):

rossf7 (approved):

rossf7:
Thank you for the 1st contribution @SRF-Audio hopefully 1st of many! 🙌 

The pipeline is triggered daily at 08:00 UTC so will check it in the morning.
https://github.com/cncf-tags/green-reviews-tooling/blob/main/.github/workflows/project-trigger.yml#L5

SRF-Audio:
@rossf7 Woohoo! I'm looking forward to more.

PR #152, opened 2025-01-07
feat: Proposal to migrate pipeline to Dagger Go SDK

### LINKED ISSUE #154: [Proposal] Migrate pipeline to Dagger Go SDK
Before the break I got a tip from @dipankardas011  about this newish k3s module for dagger
https://daggerverse.dev/mod/github.com/marcosnils/daggerverse/k3s

This got me thinking dagger could help with some pain points I have with the current pipeline

- Pipeline logic can't easily be run locally
- PRs from forked repos fail because they can't access the kubeconfig secret (for very sensible security reasons :)

I did some hacking on this over the holidays and got our pipeline running in a k3s container managed by the module.

## Benefits

- Whole pipeline including deploying falco and creating benchmark jobs can be run locally in a kind cluster
- Whole pipeline can be developed in go. GitHub Actions workflow just becomes a wrapper that calls the dagger pipeline
- Automated tests can be run with full stack (including kepler in estimation mode) in a k3s container on a GitHub Actions runner
- k3s test clusters use same flux manifests as prod cluster (just needs patches to helmrelease values for volume mounts)

## Weaknesses

- There is a learning curve for dagger
- Barrier to entry for new contributors may be slightly higher (having go and docker knowledge helps)

## Alternatives considered

`act` is another tool for running actions workflows locally but using it with kind needs a "kitchen sink" docker image that is 18Gb!
https://github.com/nektos/act/issues/335

Due to size I've split the changes into 2 PRs. #152 ports the current pipeline to dagger. #153 adds the integration test and docs for running the pipeline locally.

I did invest some time here because I think its a good solution for us. However it was also fun to learn something new and I'll probably use dagger for some of my own projects. So if we don't go this direction no worries! ☺️

rossf7:
#### What type of PR is this?

kind/feature

Migrates current pipeline logic to Dagger SDK. See #154 for why I'm proposing this.

Fixes #154

Due to size I've split the changes into 2 PRs. 

In the 2nd PR you can see the [docs](https://github.com/cncf-tags/green-reviews-tooling/blob/ef34b56ad14e662c0ec8740b65f5a5e98014f5ce/docs/dagger/README.md) for running the pipeline locally and a very basic [integration test](https://github.com/cncf-tags/green-reviews-tooling/blob/ef34b56ad14e662c0ec8740b65f5a5e98014f5ce/.github/workflows/benchmark-pipeline-test.yml) and example [run](https://github.com/cncf-tags/green-reviews-tooling/actions/runs/12613367002/job/35151146308).


Dockerfile
leonardpahlke, line 1: maybe a comment would help so the purpose of the Dockerfile is clear

dagger.json
leonardpahlke, line 1: there is a bunch of other stuff in the repo. perhaps move the dagger stuff in a subfolder? not `.`. idk the dagger conventions.

pkg/pipeline/benchmark.go
```diff
@@ -0,0 +1,193 @@
...
+	_, err = p.exec(ctx, cmd.Apply(benchmarkJobURL))
+	if err != nil {
+		return nil, err
+	}
+
+	// Wait for pods to be ready.
+	_, err = p.exec(ctx, cmd.WaitForNamespace(benchmarkNamespace))
+	if err != nil {
+		return nil, err
+	}
```
leonardpahlke, line 81: you can write these things a bit shorter fyi

if _, err := p.exec(...); err != nil {
return nil, err
}

pkg/pipeline/benchmark.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// Create benchmark job resources.
+	_, err = p.exec(ctx, cmd.Apply(benchmarkJobURL))
+	if err != nil {
+		return nil, err
+	}
+
+	// Wait for pods to be ready.
```
leonardpahlke, line 77: idk if the naming is a bit confusing. wait for pods then call wait for namespace. its fine I guess.

pkg/pipeline/benchmark.go
```diff
@@ -0,0 +1,193 @@
...
+
+func New(container *dagger.Container, dir *dagger.Directory, kubeconfig *dagger.File) (*Pipeline, error) {
+	return &Pipeline{
+		container:  container,
+		dir:        dir,
+		kubeconfig: kubeconfig,
+	}, nil
+}
+
+// Benchmark measures the sustainability footprint of CNCF projects.
```
leonardpahlke, line 35: maybe ref docs here.

pkg/pipeline/benchmark.go
```diff
@@ -0,0 +1,193 @@
...
+
+	_, err = p.delete(ctx, cncfProject, config, benchmarkJobURL)
+	if err != nil {
+		return nil, err
+	}
+
+	return p.container, nil
+}
+
+func (p *Pipeline) Terminal(ctx context.Context) (*dagger.Container, error) {
```
leonardpahlke, line 55: what's this?

pkg/pipeline/benchmark.go
```diff
@@ -0,0 +1,193 @@
...
+		return nil, err
+	}
+
+	_, err = p.execWithNewFile(ctx, fileName, fileContents, cmd.Apply(fileName))
+	if err != nil {
+		return nil, err
+	}
+
+	// Allow time for pods to be created.
+	time.Sleep(15 * time.Second)
```
leonardpahlke, line 131: why sleep? can we not "wait for namespace"? otherwise we may need to add a loop around sleep?

pkg/pipeline/benchmark.go
```diff
@@ -0,0 +1,193 @@
...
+
+	return p.container, nil
+}
+
+func (p *Pipeline) echo(ctx context.Context, msg string) (string, error) {
+	// Bust cache to ensure commands are run.
+	return p.container.WithEnvVariable("BUST_CACHE", time.Now().String()).
+		WithExec([]string{"echo", fmt.Sprintf("'%s'", msg)}).
+		Stdout(ctx)
+}
```
leonardpahlke, line 146: don't get the purpose immediately

pkg/pipeline/benchmark.go
```diff
@@ -0,0 +1,193 @@
...
+		WithEnvVariable("KUBECONFIG", "/.kube/config").
+		WithFile("/.kube/config", p.kubeconfig)
+}
+
+func getManifestPath(project, config string) string {
+	if config == "" {
+		config = project
+	}
+	return filepath.Join("/projects", project, fmt.Sprintf("%s.yaml", config))
+}
```
leonardpahlke, line 193: perhaps define some of these variables at the top of the file? and ref vars. I usually try to avoid any kind of magic string number inline.

main.go
```diff
@@ -0,0 +1,64 @@
...
+	return pipeline.New(container, source, configFile)
+}
+
+func build(src *dagger.Directory) *dagger.Container {
+	return dag.Container().
+		WithDirectory("/src", src).
+		Directory("/src").
+		DockerBuild().
+		WithMountedDirectory("/src", src).
+		WithWorkdir("/src")
```
leonardpahlke, line 52: perhaps define /src as var at the top of the file.

leonardpahlke (commented):
just browsed over it. haven't tried dagger.

leonardpahlke:
🙌

rossf7 (commented):
@leonardpahlke Thank you for the review that's great feedback. I'll add comments and address your feedback on the go code.

Some of this is due to Dagger's approach. I can also demo it at the meeting tomorrow.

pkg/pipeline/benchmark.go
leonardpahlke: we could add go linting to the ci/cd https://golangci-lint.run

dagger.json
leonardpahlke: ah, was thinking off `"source": "."` this part not the dagger.json itself

rossf7: Ah got it.

We use the manifests in the clusters and projects dirs so by making the source `.` we can keep a single copy of these files.

rossf7:
@leonardpahlke Going to close. Multiple PRs is too confusing! 😅 

Thank you for the review! Will address your feedback in the other PR.

PR #153, opened 2025-01-07
feat: Add integration test with dagger and docs

rossf7:
#### What type of PR is this?

kind/feature


docs/dagger/README.md
```diff
@@ -0,0 +1,68 @@
...
+_EXPERIMENTAL_DAGGER_RUNNER_HOST="kube-pod://$DAGGER_ENGINE_POD_NAME?namespace=dagger"
+export _EXPERIMENTAL_DAGGER_RUNNER_HOST
+```
+
+## Run pipeline
+
+- Bootstrap flux and install manifests from [/clusters/base/](/clusters/base/)
+
+```sh
+dagger call setup-cluster --source=. --kubeconfig=/src/kind-in-cluster
```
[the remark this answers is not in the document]
ChrisChinchilla: Ahhh, I see, those methods are in this PR, and not https://github.com/cncf-tags/green-reviews-tooling/issues/154

docs/dagger/README.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+- Run the pipeline and execute tests on completion
+
+```sh
+dagger call benchmark-pipeline-test --source=. --kubeconfig=/src/kind-in-cluster
```
ChrisChinchilla, line 54: I think this needs to be:

```sh
dagger call benchmark-pipeline-test \
  --source=. \
  --kubeconfig=/src/kind-in-cluster \
  --cncf-project='falco' \
  --config='modern-ebpf' \
  --version='0.39.2' \
  --benchmark-job-url='https://raw.githubusercontent.com/falcosecurity/cncf-green-review-testing/e93136094735c1a52cbbef3d7e362839f26f4944/benchmark-tests/falco-benchmark-tests.yaml' \
  --benchmark-job-duration-mins=2
```

rossf7: @ChrisChinchilla thank you and sorry for being chaotic with the PRs 😅 

Fixed 👍

locomundo (commented):
Nice work, Ross. If this works and proves to make it easier for people to work on Green Reviews project, it's going to be great.

docs/develop/README.md
```diff
@@ -0,0 +1,71 @@
...
+- `helm` https://helm.sh/docs/helm/helm_install/
+- `yq` https://github.com/mikefarah/yq/#install
+- `golang-ci` https://golangci-lint.run/welcome/install/#local-installation
+
+## Setup
+
+- Verify CLIs are installed and run Go linter.
+
+```sh
+make
```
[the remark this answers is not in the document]
rossf7: @locomundo thank you for trying the steps. The problem is because dagger uses code generation in the internal dir but the recommendation is not to store in git

https://github.com/cncf-tags/green-reviews-tooling/blob/553898dc9b7b05ead8342bb43bcf6832b05db92b/.gitignore#L6-L8

For now I've removed the linter as I think its more important to try the dagger steps.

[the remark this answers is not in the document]
rossf7: @locomundo there was another problem with the `make install` command.

On first run the pod name was missing from the 2 dagger env vars.  It should be OK now but if not try running `make install` again.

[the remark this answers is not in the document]
locomundo: super, that fixed it, thanks!!!

locomundo:
The installation works for me now, lovely

.github/workflows/benchmark-pipeline-test.yml
dipankardas011, line 1: This is for testing it out?

pkg/pipeline/benchmark.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	if _, err = p.execWithNewFile(ctx, fileName, fileContents, cmd.Apply(fileName)); err != nil {
+		return nil, err
+	}
+
+	// Allow time for pods to be created. This is needed because there is a
+	// delay while flux reconciles the manifest. Without it the following
+	// kubectl wait command will fail.
+	time.Sleep(15 * time.Second)
```
dipankardas011, line 139: Is this sleep interval a random guess or there is something we are referring to, like any docs?

pkg/pipeline/setup_cluster.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	return parts[0], nil
+}
+
+// clusterManifests are applied to bootstrap the monitoring stack.
+func clusterManifests() []string {
+	return []string{
+		// Namespace must be created first for dependencies.
+		"/clusters/base/monitoring-namespace.yaml",
```
dipankardas011, line 83: I think we should use filepath package?

dipankardas011 (approved):
lgtm

.github/workflows/benchmark-pipeline-test.yml
rossf7: @dipankardas011 Yes, it runs the pipeline with these params. Here is an example [run](https://github.com/cncf-tags/green-reviews-tooling/actions/runs/13161126733/job/36729899352?pr=153)

At the moment it just checks there are no errors. Once the pipeline writes the JSON file we could also test it has the correct data format.

docs/develop/README.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+- Create kind cluster and add kubeconfig to source dir so dagger can access it.
+
+```sh
+kind create cluster
+kind get kubeconfig | yq e '.clusters[0].cluster.server = "https://kubernetes.default"' - > green-reviews-test-kubeconfig
```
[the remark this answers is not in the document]
rossf7: @dipankardas011 I changed this to `yq eval` as that should work for both linux and mac. Does it work for you?

go.mod
```diff
@@ -0,0 +1,51 @@
...
+	google.golang.org/protobuf v1.35.2 // indirect
+)
+
+replace go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc => go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc v0.0.0-20240518090000-14441aefdf88
+
+replace go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp => go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp v0.3.0
+
+replace go.opentelemetry.io/otel/log => go.opentelemetry.io/otel/log v0.3.0
+
+replace go.opentelemetry.io/otel/sdk/log => go.opentelemetry.io/otel/sdk/log v0.3.0
```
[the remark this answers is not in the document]
rossf7: These are from `dagger develop` command not me :wink:
https://docs.dagger.io/reference/cli/#dagger-develop

I think they are needed for code gen

pkg/pipeline/benchmark.go
rossf7: This is basically the same as we have in the current pipeline.
https://github.com/cncf-tags/green-reviews-tooling/blob/87bbf10320f714ae88e74312dce5f4419239e5aa/.github/workflows/benchmark-pipeline.yaml#L104

It's not pretty but pipeline will fail without it and we spent quite some time debugging it.

pkg/pipeline/setup_cluster.go
rossf7: What is the benefit for using filepath? We just kubectl apply each string.

docs/develop/README.md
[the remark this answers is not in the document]
dipankardas011: OKay both of them works for me

go.mod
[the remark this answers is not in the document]
dipankardas011: Interesting, then its good to go, fun part is it replaces but it also doesn't seems to replace 🤔

pkg/pipeline/benchmark.go
dipankardas011: okay

pkg/pipeline/setup_cluster.go
dipankardas011: If this command is going to run always in linux environment then we don't need any changes

go.mod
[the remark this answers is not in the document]
rossf7: Hmm that is interesting. To investigate a bit further I upgraded dagger to latest and re-ran the develop command.
This updated the go.mod including the replaces.

Let's go with this but keep an eye on it for future dagger upgrades.

pkg/pipeline/setup_cluster.go
rossf7: Good to check on support for other OSes. I think this command is always going to run in a linux container so I think we're good. We can revisit later if it becomes a problem.

main.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// newPipeline creates a new benchmark pipeline.
+func newPipeline(ctx context.Context, source *dagger.Directory, kubeconfig string) (*pipeline.Pipeline, error) {
+	var configFile *dagger.File
+	var err error
+
+	container := build(source)
+
+	if kubeconfig == "" {
```
locomundo, line 114: I didn't try this path when the cluster doesn't exist yet. I just wonder if to reduce complexity we can just fail if the cluster was not created beforehand, so there is only one happy flow.
Because does it mean that in one case we use kind to start a cluster and in the other k3s?

But take it with a salt of grain because I'm still learning how this all works with Dagger :).

We can start working with it and see how it goes.

pkg/pipeline/benchmark.go
```diff
@@ -0,0 +1,211 @@
...
+		return nil, err
+	}
+
+	if _, err = p.execWithNewFile(ctx, fileName, fileContents, cmd.Apply(fileName)); err != nil {
+		return nil, err
+	}
+
+	// Allow time for pods to be created. This is needed because there is a
+	// delay while flux reconciles the manifest. Without it the following
+	// kubectl wait command will fail.
```
locomundo, line 138: Suggestion to add an echo "Waiting 15 seconds for pods to be created" or "Waiting a few seconds for the deployment to complete"

```
p.echo(ctx, "Waiting 15 seconds to let deployment complete")
```

in case the 15 seconds are not enough then we have a clue that this could be the reason... Always tricky to wait for a fixed amount of time

locomundo (approved):
It looks great!

I added a couple of comments, but I might be confused. It it's up to me we are good to move forward.

Very nice work 👍

nikimanoledaki (approved):
LGTM!! The only thing I wasn't sure about is whether we should move `main.go` into its own dir so that we can have multiple Go applications in the same repo to have a structure that makes room for the Go client that'll be added soon as well 🤔

main.go
rossf7: > I didn't try this path when the cluster doesn't exist yet. I just wonder if to reduce complexity we can just fail if the cluster was not created beforehand, so there is only one happy flow.

@locomundo It would be good to reduce complexity but we do use both modes.

If there is a kubeconfig we assume prometheus and kepler are already deployed using flux. This is the case when running against the equinix cluster. For local dev it means the setup code only has to be run once and then the pipeline can be run multiple times.

> Because does it mean that in one case we use kind to start a cluster and in the other k3s?

Yes, this isn't ideal. We could use k3d instead of kind so we always use k3s? 

I went with kind because its probably more widely used. However switching to k3d shouldn't be too big. It'd mainly involve testing and updating the docs.

pkg/pipeline/benchmark.go
rossf7: @locomundo Great idea I've added this extra logging.

I agree using a fixed wait time is problematic. We could actually use `flux reconcile kustomization falco-cncf-green-review-testing` as an alternative but it would be project specific.

A possible solution is adding the flux resources to the metadata in `projects.json`. WDYT?

rossf7:
> The only thing I wasn't sure about is whether we should move main.go into its own dir so that we can have multiple Go applications in the same repo to have a structure that makes room for the Go client that'll be added soon as well 

@nikimanoledaki Yes, I'm not sure on this either. There are some benefits to having it in the root as in the dagger container it means we re-use the manifests in the `clusters` and `projects` dirs.

For the go client code I was thinking we could add packages like `pkg/collect` and then call them from the pipeline. WDYT?

I'd say lets keep an eye on this as we add more complex logic and we can restructure if necessary.

PR #157, opened 2025-02-06
added governance document proposal

AntonioDiTuri:


locomundo:
Nice, @AntonioDiTuri , this clarifies the process with tracking issues and actions very well.

I was thinking if, in the interest of agility like we were discussing in the last meeting, we could simplify it. 

What if we only had feature request issues, and these can change state if they are chosen to be worked on? So no tracking issues or action issues. Would that make sense or are we missing a lot?

If it's simpler we will have less administrative work and it will be clear for newcomers, so we will go faster...

Just a suggestion, perhaps nobody agrees :)

rossf7:
@AntonioDiTuri this is nice and the diagram makes it very clear

> What if we only had feature request issues, and these can change state if they are chosen to be worked on? So no tracking issues or action issues. Would that make sense or are we missing a lot?

@locomundo I like this idea as I also think it would reduce overhead. 

Maybe we could use labels for the state and add a `feature_request` label via the issue template? We could change the label to `feature` when its assigned a release.

docs/governance/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+The working group follows the Roadmap that can be find in the root Readme. 
+
+1. At any point in time any person can submit a feature request via issue
+2. Every quarter the working group aims at releasing a new project version
+3. At the beginning of every quarter the working group engages in a formal discussion to decide which features should be included in the quarter release
+4. Every feature is mapped into a TRACKIUNG issue monitoring the status of the feature
+5. Every tracking issue will have a set of ACTION issues that will detail the steps needed to build the feature
+6. ACTION issues can be marked as `good first issues` for beginner friendly contributions
+7. Each ACTION issue will be closed after the submission of a formal PR approved by the wg-leads
```
nikimanoledaki, line 14:
```suggestion
5. Feature requests can have [sub-issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/adding-sub-issues) that may provide more detail for the steps needed to build the feature
6. Feature requests should have enough detail so that they can be marked as `good first issues` for new contributors
7. Each feature request will be closed after the submission of a formal PR approved by the wg-leads
```

docs/governance/README.md
```diff
@@ -0,0 +1,18 @@
+# Governance
```
nikimanoledaki, line 1: `Governance` has a different meaning IMO in this context - it tends to mean the organizational structure of a OSS project, like the leads/maintainers. I'd say this is more about the planning process, in agile terms.

```suggestion
# Planning Features
```

docs/governance/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+This document attempts to describe the governance framework adopting for contributing to this repo.
```
nikimanoledaki, line 3:
```suggestion
This document describes the feature planning process for contributing to this repo.
```

nikimanoledaki (commented):
I left some suggestions based on previous conversations, hope this is helpful :)
I agree that if we add `Feature request` issue templates, this will help us a lot. I was looking into this today. It'll be easier than introducing a new terminology such as tracking/action issues. Hopefully we will have less of a need for tracking issues if we improve the way we go over our backlog etc. Lastly, using sub-issues should be enough to track specific implementation details for feature requests.

AntonioDiTuri:
@locomundo @rossf7 @nikimanoledaki addressed the comments and simplified the diagram, let me know if something else needs to be adjusted :)

nikimanoledaki (approved):
LGTM!! Thanks @AntonioDiTuri!

PR #162, opened 2025-02-13
Update Roadmap process in README.md

nikimanoledaki:
#### What type of PR is this?

kind/documentation

Updating the roadmap planning process to be more flexible / agile and less waterfall.
Now that we have a board, we can add issues to it directly: https://github.com/orgs/cncf-tags/projects/5/views/7


rossf7 (approved):
LGTM

Board will be easier to keep up to date 👍

PR #165, opened 2025-02-25
fix: Support latest falco release and fixes for integration test

rossf7:
#### What type of PR is this?

kind/bug

There are some breaking changes in the latest Falco release (0.40.0) we need to integrate.
See https://falco.org/blog/falco-0-40-0/#breaking-changes-and-deprecations

~~Needs https://github.com/falcosecurity/cncf-green-review-testing/pull/23 to be merged first.~~ EDIT: Merged

I also made some fixes for the integration test that I missed in my previous PR. I'll add comments for those.

### DIFF
--- a/.github/workflows/benchmark-pipeline-test.yml
+++ b/.github/workflows/benchmark-pipeline-test.yml
@@ -17,8 +17,3 @@ jobs:
           version: "0.15.4"
           args: benchmark-pipeline-test
               --source='.'
-              --cncf-project='falco'
-              --config='modern-ebpf'
-              --version='0.39.2'
-              --benchmark-job-url='https://raw.githubusercontent.com/falcosecurity/cncf-green-review-testing/2551137b1a09bd0594f76b09e82e08c98f95efd3/benchmark-tests/falco-benchmark-tests.yaml'
-              --benchmark-job-duration-mins=2
--- a/.github/workflows/benchmark-pipeline-test.yml
+++ b/.github/workflows/benchmark-pipeline-test.yml
@@ -3,9 +3,6 @@ name: Benchmark Pipeline Test
 on:
   push:
 
-concurrency:
-  group: benchmark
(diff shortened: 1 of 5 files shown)


.github/workflows/benchmark-pipeline-test.yml
```diff
@@ -3,9 +3,6 @@ name: Benchmark Pipeline Test
 on:
   push:
 
-concurrency:
-  group: benchmark
-
```
rossf7, line 8: I'm afraid this was copy pasta from the benchmark pipeline workflow.

It is needed there so we only benchmark a single configuration at a time on the equinix cluster. The integration test uses a k3s container so this isn't needed.

.github/workflows/benchmark-pipeline-test.yml
```diff
@@ -17,8 +14,3 @@ jobs:
           version: "0.15.4"
           args: benchmark-pipeline-test
               --source='.'
-              --cncf-project='falco'
-              --config='modern-ebpf'
-              --version='0.39.2'
-              --benchmark-job-url='https://raw.githubusercontent.com/falcosecurity/cncf-green-review-testing/2551137b1a09bd0594f76b09e82e08c98f95efd3/benchmark-tests/falco-benchmark-tests.yaml'
-              --benchmark-job-duration-mins=2
```
rossf7, line 24: These params are duplicated in the BenchmarkPipelineTest job we run when calling `make test`
https://github.com/cncf-tags/green-reviews-tooling/blob/56fedac11a85ea3231c02568693832b9c245b7e5/main.go#L39-L57

I think its best to use the default values there and not duplicate here as they will get out of sync.

nikimanoledaki (approved):

nikimanoledaki:
Thank you for these bug fixes @rossf7, merged! 🚀

PR #170, opened 2025-03-11
feat: Use Oracle Cloud bare metal instance

### LINKED ISSUE #166: Investigate using OCI bare metal for green reviews cluster
Currently the green reviews cluster consists of physical servers from Equinix Metal using credits donated to CNCF.

Equinix Metal is being sunset in June 2026 and we're investigating whether we can use Oracle Cloud Infrastrucure (OCI) as an alternative.

## Goals

- Continue using bare metal servers with RAPL enabled for collecting energy metrics
- Retire Equinix Metal servers and user accounts
- Select OCI region primarily powered by renewable energy
- Reduce footprint of green reviews cluster where possible

## Non-Goals

- Provision servers on demand (will be addressed in #67)

## Region Selection

Currently our OCI account only has access to US East. Provisioning servers in AMS has a higher renewable energy % and a lower PUE which we include in the [SCI calculation](/docs/measurement/sci.md).

CDG is an alternative but it has a higher PUE and a lower average carbon intensity due to the high % of nuclear energy in the French grid.

Netherlands means we will use a more typical value for carbon intensity in the SCI calculation.

| REGION | CODE | RE% | PUE |
|--------|------|-----|-----|
| US East (Ashburn) | IAD | 93% | 1.34 |
| North West Netherlands (Amsterdam) | AMS | 100% | 1.23 |
| France Central (Paris) | CDG | 100% | 1.30 |

Source: https://www.oracle.com/a/ocom/docs/corporate/citizenship/clean-cloud-oci.pdf

## Bare Metal Instances

OCI bare metal instances have access to RAPL and are compatible with our stack (ubuntu 24.04 / k3s / kepler).

Unlike Equinix Metal there is no ACPI power monitor so kepler cannot measure idle power but this is not critical since we are primarily interested in measuring dynamic power.

OCI use the term Shape for instance types. The smallest bare metal shape is sufficient for our current use case.

| Shape | OCPU | Memory (GB) | Local Disk | Max Network Bandwidth | Max VNICs Total: Linux | Max VNICs Total: Windows |
|-------|------|-------------|------------|----------------------|------------------------|-----------------------------|
| BM.Standard3.64 | 64 | 1024 | Block storage only | 2 x 50 Gbps | 256 | 129 (1 on the first physical NIC, 128 on the second) |

https://docs.oracle.com/en-us/iaas/Content/Compute/References/computeshapes.htm#bm-standard

## VM Instances

The benchmarking node will continue to be a physical server to have access to RAPL.

For the control plane and the internal node with flux and prometheus we could use virtual machines including burstable instances since a physical server is not necessary.

https://docs.oracle.com/en-us/iaas/Content/Compute/References/burstable-instances.htm

This option is not possible with Equinix Metal. The downside would be that energy measurements for components running on these nodes would be in estimation mode.

rossf7:
#### What type of PR is this?

kind/feature

Fixes 

WIP on Equinix to Oracle migration


infrastructure/oracle-cloud/main.tf
```diff
@@ -0,0 +1,186 @@
+terraform {
+  required_providers {
+    equinix = {
```
locomundo, line 3:
```suggestion
    oci = {
```

infrastructure/oracle-cloud/main.tf
```diff
@@ -0,0 +1,186 @@
...
+  # Allow inbound API server traffic
+  ingress_security_rules {
+    protocol  = "6" # TCP
+    source    = "0.0.0.0/0"
+    stateless = false
+
+    tcp_options {
+      min = 6443
+      max = 6443
+    }
```
locomundo, line 82: just wondering how we use access to 6443 and if it's necessary to open it to the world. you probably have a good reason, I just don't know :)

locomundo (commented):
Nice, Ross! Are we also going to have a github action to plan and deploy it?

rossf7:
> Are we also going to have a github action to plan and deploy it?

Thanks @locomundo I posted update here https://github.com/cncf-tags/green-reviews-tooling/issues/166#issuecomment-2715232854

rossf7:
Closing in favour of #170

PR #171, opened 2025-03-17
Update contribution docs

### LINKED ISSUE #90: [DOCUMENTATION] Revise content in falco.md
Comment from Leo in #89 regarding content in `falco.md` file:

> what is the idea of this section? We could put a list of cncf projects that are integrated with link to pipelines + link to project config. So one file with a table of projects? This file reads like a blog post. I do not expect that we will have the same depth of analysis for future projects past falco.

nikimanoledaki:
closes #90 

#### What type of PR is this?

kind/documentation

* Added an issue template for CNCF Projects to register their interest in the project
* Added and sorted out docs
* Removed an issue template that was added by accident


nikimanoledaki:
re: pipeline CI test, I wonder if we could edit the workflow to only run when there are changes to specific repos that touch the code itself rather than all pushes to `main`

locomundo (approved):
Looking good!!

PR #184, opened 2025-06-13
fix: Trigger pipeline weekly on Wednesdays

rossf7:
#### What type of PR is this?

kind/enhancement

Quick win to run pipeline once a week on Wednesday at 08:00 UTC instead of daily.

This is so we consume less Oracle Cloud resources.


locomundo (approved):
Looks good to me :)

nikimanoledaki (approved):

PR #188, opened 2026-03-18
[DRAFT] issue #161 to compute SCI using prom data

### LINKED ISSUE #161: Compute the SCI through the Prometheus client
Compute the SCI through the Prometheus client.

We can [run the PromQL query](https://github.com/cncf-tags/green-reviews-tooling/blob/45185e85d0c8e7d8a06009b306d1421ace9c749a/clusters/base/falco-sci.yaml#L478) which is already available using the Prometheus Go Client

davemasselink:

### DIFF
--- a/pkg/cmd/cmd.go
+++ b/pkg/cmd/cmd.go
@@ -1,6 +1,9 @@
 package cmd
 
-import "fmt"
+import (
+	"fmt"
+	"net/url"
+)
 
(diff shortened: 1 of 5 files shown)


davemasselink:
@AntonioDiTuri, have a quick SCI compute approach sketched out here. I'm about to now install dagger and test in full integration suite.

